PR #24, opened 2023-02-02
feature / 21 new iaudio

### LINKED ISSUE #21: Give easy access to AudioCues in A singleton
Something similar to the work done for #19 , but for playing sounds and music:

```
A.sfx_toy_car.play()
A.mx_main_theme.play_no_run()
```

There is already progress on this in `popochiu-dev/new_audio_manager` branch, but it will require to rework the **AudioManager** so we don't have conflicts with old projects (including the one used for the tutorials).

From what I found, the **AudioManager** shouldn't be loaded as an autoload. It should work like the **GraphicInterface** and **TransitionLayer** scenes, which are instantiated and added to the tree in runtime. Then we'll need an **IAudio** script which will contain all the functions that are currently part of the **A** singleton, i.e. `A.play(...)`, `A.play_music(...)`, `A.stop(...)`, and so on. After that we can have our new **A** script populated by code with the references to all the **_AudioCues_** in the project.

> I'll move my changes from `popochiu-dev/new_audio_manager` to `feature/21-new_audio_manager` in order to start working on this.

mapedorr:
This closes #21 by updating the way the AudioManager works.

Now it is loaded directly as a child of `Popochiu.tscn`, instead of being the audio autoload. For that, a new script was created: **IAudio.gd** (following the pattern of other singletons). In that script are now the methods to play and stop audio cues (those colled by `A.play(...) | A.stop(...) | ...`. A new script **A.gd** is created when the Engine loads and in that script the AudioCues in the project are stored as variables to allow autocompletion by doing `A.mx_main_theme.play() | A.sfx_toy_car.play_no_run() | ...`.

In projects, previous to **v1.9.0**, with audio files already set, the plugin will update those during load so they are now either an AudioCueSound or an AudioCueMusic. This is required to have autocompletion in **A.gd**.

**A.gd** will be the script to autoload with the **A** shortcut.

FIX: Now devs can show inline dialogs while playing a branching dialog.


addons/Popochiu/Engine/AudioManager/AudioCue.gd
```diff
@@ -5,7 +5,7 @@ export var audio: AudioStream
 export var loop := false setget _set_loop
 export var is_2d := false
 export var pitch := 0.0
-export var volume := 1.0
+export var volume := 0.0
```
stickgrinder, line 8: Hah, I like what you did here!

addons/Popochiu/Engine/Interfaces/IAudio.gd
```diff
@@ -0,0 +1,119 @@
...
+
+func change_cue_pitch(cue_name: String, pitch := 0.0) -> void:
+	E.am.change_cue_pitch(cue_name, pitch)
+
+
+func change_cue_volume(cue_name: String, volume := 0.0) -> void:
+	E.am.change_cue_volume(cue_name)
+
+
+func semitone_to_pitch(pitch: float) -> float:
```
stickgrinder, line 118: Unsure, but isn't this just a duplicate? Maybe the audiomanager one can be reused (of viceversa?)

addons/Popochiu/Engine/Popochiu.gd
```diff
@@ -62,8 +63,12 @@ onready var _saveload := SaveLoad.new()
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ GODOT ░░░░
 func _ready() -> void:
+	if Engine.editor_hint: return
```
stickgrinder, line 66: What are you doing here? Returning when you are running the editor I guess, but why?

stickgrinder:
LGTM

Issue #18, opened 2023-01-27
Add native animation support to PopochiuCharacter


stickgrinder:
_This issue is somehow related to #8_

The current implementation of each game object only supports a sprite to represent the object in both the editor scene and the game.
Also, besides flipping the player character's sprite when it moves, no accurate directions are taken into account to turn the sprite accordingly to the movement direction.

## What should be done

* [x] Add an AnimationPlayer element to PopochiuCharacter base scene to control the sprite by means of one or more animations
* [x] Make sure the character actually plays the walk animation while walking
* [x] Take the 8 directions that are listed in the `LOOKING` enum into consideration when an animation is played (i.e. if the character is facing down-right and we have an animation to make them talk down-right, that has to be played.
* [x] Make sure to fall back on the 4 directions or the most simple case of "no direction" (like there is no "talk right" but just "talk" so use it, flipping the animation if necessary)
* [x] Introduce a naming convention for directional animations (eg `<animation_name>_dl` for down-left, `<animation_name>_d` for down, `<animation_name>_l` for left, and so on.

Maybe I'm overlooking something meaningful, in case I'll take more notes in here.

stickgrinder:
This has been long closed.

Issue #8, opened 2022-09-01
Aseprite importers


stickgrinder:
This idea is heavily inspired by PowerQuest's importers, which is a key component of the framework's streamlined workflow.

Basically, when a new room or character are created, a new importer prefab is generated. This importer takes a properly organized Aseprite file as source and automatically creates animations for room props, character or inventory items.

## Aseprite file layout

The Aseprite file for a room will contain all the room props isolated on their own frames. Each prop will have its frame(s) tagged wiith the prop's name. Some props will be represented by a single frame, some other will have more than one frame.

The following example shows a quick and dirty scene with a `Background` prop, a `Trees` prop, a `Tent` prop and a `Rock` prop. All of them are static, represented by a single frame. The `Fire` prop instead is animated so its tag extends along four frames.

![immagine](https://user-images.githubusercontent.com/625436/188014204-b732d27d-5add-4ddf-ac08-60bd0b78233d.png)

Layers, although convenient, are not relevant to the file organization. The first frame contains the whole image but is not tagged. This frame is also there for convenience but will not be taken into consideration by the importer, lacking a named tag.

The props frames will have a transparent background, as shown in this animation:

![Popochiu Room Example](https://user-images.githubusercontent.com/625436/188015623-02973c8c-d811-4505-8f7a-aa927670f446.gif)

## Importer

Each specific room should ideally have an importer panel. This panel may be organiized as follows:

![importer-interface-mockup](https://user-images.githubusercontent.com/625436/188019083-1af081dc-fb00-4807-8f3b-b62f03bdf2f7.jpg)

1. Here you specify the source Aseprite file
2. This button scans the file for tags and their position/duration. Clicking it populates the subsequent list
3. Each tag found is shown with name (it will become the animation name), starting frame (for convenience) and the following actions:
  * Open the animation in Godot Animation Editor
  * Reload animation sprites and rebuild the animation
  * Toggle animation loop (not all animations are looping, this should ideally set/unset the animation flag without the need to open the animation in the editor, as well as showing if the animation is looping at a glance)
  * Additional dropdown menu for possible other actions (delete all frames+anim, rename, whatever)
4. This will bulk reimport all sprites, useful for massive operations like global recoloring, palette changes, repainting over progammer's art, etc
5. This will rebuild all animations, useful after re-scanning a file, if the animations changed number of frames

[...]

PR #25, opened 2023-02-04
refs #18: 

mgdesign:
Added walk animation logic in all directions, with updated the main character sprite. 
The room script is in charge of both moving and turn the character durign movement.

Done:
- character facing direction switch logic based on destination point 
- character animation based on AnimationPlayer and naming conventions 
- character turning and animation triggers while moving through the walk cycle
- public play_animation() on the singleton, that can also be used in other interactions
- animation added to the AnimationPlayer for any orientation are now play according to the player looking point
- basic fallback logic on walk animations
- added talk animation to test using the simplest case of animation 
- updated demo game main character sprite to introduce animated walk up/down/left cycles
- add the AnimationPlayer in the character template

TODO:
- figure out if it is possible to implement the logic that relies on character looking direction and fallback for all the animations in general


mapedorr:
This has been merged to main along with the changes for #8

PR #50, opened 2023-04-23
refs #18 - Fix direction finding bug

stickgrinder:
Added proper fallbacks to every direction so that animations can't be missedif there is at least a directional or generic one.


mapedorr (approved):
Everything looks good.

Issue #46, opened 2023-04-10
Add animation-control functions to Characters and other objects


stickgrinder:
This has been split from #18 to avoid scope-creep and since the best option is to have those enhancements in 2.0.

We have to enhance the interface of the PopochiuCharacter, PopochiuProp and PopochiuInventory classes to introduce more animation-controlling methods.  
We have at least to provide:

* [x] `pause_animation`
* [x] `resume_animation`
* [x] `stop_animation`
* [x] (optional) a non-blocking version of `play_animation` that works in the background
* [x] ~~(optional) a `wait_for_idle` function that blocks a script until a character is not back to idle animation~~ **see comments**

Of course the whole thing, plus conventions must be documented in the wiki.

stickgrinder:
`wait_for_idle` does not belong here. It is not meant to wait for animations alone but for movement or talking etc.

It is worth adding it as an optional to have better control during cutscenes.

PR #60, opened 2023-06-21
refs #46: extended animation API for characters

stickgrinder:
This simple one introduces three useful functions.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -324,6 +324,18 @@ func queue_play_animation(animation_label: String, animation_fallback := 'idle',
 	return func(): await play_animation(animation_label, animation_fallback)
 
 
+func queue_stop_animation():
+	return func(): await stop_animation()
+
+
+func queue_pause_animation():
+	return func(): pause_animation()
+
+
+func queue_resume_animation():
+	return func(): resume_animation()
+
+
 func play_animation(animation_label: String, animation_fallback := 'idle'):
 	if not has_node("AnimationPlayer"):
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -346,6 +358,25 @@ func play_animation(animation_label: String, animation_fallback := 'idle'):
...
 
+func stop_animation():
+	# TODO: consider returning if idle is playing, but in a way
+	# that's directions and animation_prefix aware!
+	var animation = $AnimationPlayer.get_animation($AnimationPlayer.current_animation)
+	var animation_loop_mode = animation.loop_mode
+	animation.loop_mode = Animation.LOOP_NONE
+	await $AnimationPlayer.animation_finished
+	_play_idle()
+	animation.loop_mode = animation_loop_mode
```
mapedorr, line 369: What would happen if one wants to stop an animation that is not on loop? Shouldn't this function call $AnimationPlayer.stop()?

mapedorr (commented):
I just have a question about the `stop()` function. Otherwise this is ready for merge.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: I was waiting for this question :D

I intended this function to stop the animation organically, when a cycle is over, going back to the idle state. This does not abruptly stop an animation midway through. If a long non-looping animation is running, this function basically does nothing.

Now that I think of it I can just test this case...

stickgrinder: One more thing is, we can parametrize this function to extend its behavior to stop the animation suddenly, no matter the state.

Issue #17, opened 2022-12-06
Sometimes Popochiu doesn't load objects data on start


mapedorr:
Opening a project sometimes doesn't show the game objects (rooms, characters, inventory items, dialogs) in the Main tab. They appear until one opens the Room tab and goes back to the Main tab.

mapedorr:
Fixed with https://github.com/mapedorr/popochiu/pull/63

Issue #58, opened 2023-06-16
Creating Props doesn't associate the .tscn properly


mapedorr:
Despite now Popochiu creates a `.tscn` file for each **PopochiuProp**, clicking the button that opens the prop's scene in the SceneTree goes to `popochiu_prop.tscn` instead of opening the scene file created for the prop.

mapedorr:
Fixed with https://github.com/mapedorr/popochiu/pull/63

Issue #56, opened 2023-06-03
Main scene changes automatically


mapedorr:
When a new room is created, it automatically becomes the main scene.

mapedorr:
Fixed with https://github.com/mapedorr/popochiu/pull/63

Issue #55, opened 2023-06-03
Can't create more than one character


mapedorr:
Once a character is created one cannot create more without restarting the engine.

mapedorr:
Fixed with https://github.com/mapedorr/popochiu/pull/63

PR #63, opened 2023-06-24
Fixes: #17 #55 #56 and #58

mapedorr:


addons/popochiu/editor/popups/create_character/create_character.gd
```diff
@@ -49,7 +49,7 @@ func _create() -> void:
 	# ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 	# Create the state Resource for the character and a script so devs
 	# can add extra properties to that state
-	var state_template: Script = load(CHARACTER_STATE_TEMPLATE)
+	var state_template: Script = load(CHARACTER_STATE_TEMPLATE).duplicate()
```
stickgrinder, line 52: I see what happens here, but to learn something solid... I guess you're working around the bug where you can create only one character?

addons/popochiu/editor/popups/create_prop/create_prop.gd
```diff
@@ -75,7 +75,7 @@ func _create() -> void:
 	prop.clickable = _interaction_checkbox.button_pressed
 	prop.cursor = Constants.CURSOR_TYPE.ACTIVE
 	
-	if _new_prop_name in ['Bg', 'Background']:
+	if _new_prop_name in ['bg', 'background']:
```
stickgrinder, line 78: Wait! Are we going to change PascalCase in node names?

addons/popochiu/editor/popups/create_prop/create_prop.gd
```diff
@@ -93,10 +93,13 @@ func _create() -> void:
 	
 	# ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 	# Add the prop to its room
-	_room.get_node('Props').add_child(prop)
+	# Instancing the created .tscn file fixes #58
+	var prop_instance: PopochiuProp = load(_new_prop_path + '.tscn').instantiate()
 	
-	prop.owner = _room
-	prop.position = Vector2(
+	_room.get_node('Props').add_child(prop_instance)
```
stickgrinder, line 99: Not blocking for this PR but maybe this logic belongs to the Room? So it can expose the logic to add a new prop to itself?

addons/popochiu/editor/popups/create_prop/create_prop.gd
```diff
@@ -122,8 +125,7 @@ func _create() -> void:
 	# Abrir las propiedades de la prop creada en el Inspector
 	_main_dock.fs.scan()
 	await get_tree().create_timer(0.1).timeout
-	_main_dock.ei.edit_node(prop)
-	
+	_main_dock.ei.edit_node(prop_instance)
```
stickgrinder, line 128: Does this focus the scene tree on the node?

addons/popochiu/editor/popups/create_room/create_room.gd
```diff
@@ -117,7 +117,8 @@ func _create() -> void:
 	
 	# ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 	# Create the room instance
-	var new_room: PopochiuRoom = preload(BASE_ROOM_PATH).instantiate()
+	var new_room: PopochiuRoom = load(BASE_ROOM_PATH).instantiate()
```
stickgrinder, line 120: Can you explain like I'm 5 what's the deal between load and preload in this context? I can't fathom the impacts.

addons/popochiu/engine/objects/graphic_interface/graphic_interface.tscn
```diff
@@ -45,6 +45,7 @@ offset_left = -8.0
 offset_top = -8.0
 offset_right = 8.0
 offset_bottom = 8.0
+grow_horizontal = 2
```
stickgrinder, line 48: What is this for?

addons/popochiu/engine/templates/hotspot_template.gd
```diff
@@ -13,8 +13,8 @@ func _on_click() -> void:
 	# For example you can make the character walk to the Hotspot and then say
 	# something:
 #	E.queue([
-#		C.walk_to_clicked(),
-#		C.face_clicked(),
+#		C.queue_walk_to_clicked(),
```
stickgrinder, line 16: That's OK but I would take a note to rewrite templates using the simpler "await" strategy. What do you think?

addons/popochiu/popochiu_plugin.gd
```diff
@@ -57,6 +57,7 @@ func _init():
 func _enter_tree() -> void:
 	if _is_first_install: return
 	
+	# Good morning, starshine. The Earth says hello.
```
stickgrinder, line 60: :D

addons/popochiu/popochiu_plugin.gd
```diff
@@ -102,7 +103,7 @@ func _enter_tree() -> void:
 	
 	# Fill the dock with Rooms, Characters, Inventory items, Dialogs and
 	# AudioCues
-	main_dock.grab_focus()
+	main_dock.call_deferred('grab_focus')
```
stickgrinder, line 106: Why do you need to defer this call here?

stickgrinder (commented):
Left some comments and questions (out of genuine curiosity). Seems good anyway.

addons/popochiu/editor/popups/create_character/create_character.gd
mapedorr: Yes. The problem was that on the process of creating our game objects the original template file was being overwritten (something that was not a problem in Godot 3.x). I forgot to put a comment to clarify what the change is intended for.

addons/popochiu/editor/popups/create_prop/create_prop.gd
mapedorr: Nope. Node names should be in PascalCase. The point of doing this here is that the name of the prop is changed to snake case when devs click the confirmation button in the popup. I can change this comparison to the `_pascal_name` variable if we want to keep consistency with Popochiu 1.x.

addons/popochiu/editor/popups/create_prop/create_prop.gd
mapedorr: Since the creation of props implies the creation of folders and files in the system, I took the decision to separate the functionality from the Room (which should handle the logic related to in-game actions). But I'm open to discuss this, of course.

addons/popochiu/editor/popups/create_prop/create_prop.gd
mapedorr: Yup.

addons/popochiu/editor/popups/create_room/create_room.gd
mapedorr: `load` tries to load the resource during runtime, `preload` does it while in the Editor. Looks like changes in Godot 4 make preloading resources behave different than in Godot 3.x. Then, preloading this audio cues here was the reason why calling something like `A.sfx_xxx.play()` didn't work. I don't know the details of that change.

addons/popochiu/engine/objects/graphic_interface/graphic_interface.tscn
mapedorr: To define how a Control node should grow. From Godot docs:

> The control will grow in both directions equally to make up if its minimum size is changed to be greater than its current size.

addons/popochiu/engine/templates/hotspot_template.gd
mapedorr: Yes... I was thinking about that too... I'll change those helper comments to the new `await` approach. Thanks for the suggestion!

addons/popochiu/popochiu_plugin.gd
mapedorr: I probably did this as an attempt to solve #17 . But what really made the trick at the end was making a deferred call to `maind_dock.fill_data()`.

addons/popochiu/editor/popups/create_prop/create_prop.gd
stickgrinder: Well, it makes perfect sense actually. Then, I would move all the creation/deletion logic to a utility class (a factory or repository). That way we can refer to that class everywhere we are trying to automatize those operations since they can become pretty complex and require context validation.

What do you think?

addons/popochiu/editor/popups/create_prop/create_prop.gd
stickgrinder: TIL how to do this and it was something I wanted to do in the importer :D Thanks!

addons/popochiu/editor/popups/create_room/create_room.gd
stickgrinder: Strange. I will dig deeper into this, seems quite impactful and important to understand.
Thanks

addons/popochiu/engine/objects/graphic_interface/graphic_interface.tscn
stickgrinder: Ah, yeah, I'm dumb! Sorry

addons/popochiu/popochiu_plugin.gd
stickgrinder: ... that I spotted below. OK then, I guess you could leave this deferred. This is another topic I need to grasp better, I admit.
Thanks for your answer anyway.

addons/popochiu/editor/popups/create_prop/create_prop.gd
mapedorr: Yes. Indeed, a thing I want to refactor is the creation popups. At the moment some of them perform similar operations (in terms of creating folders and files) but the logic is copied here and there. Now that you'll need some of those things for the importer, having that logic in a separate script will make our lives easier, and the project better structured.

PR #72, opened 2023-07-25
Add optional offset to walk_to methods

### LINKED ISSUE #69: Add offset parameter to PopochiuCharacter.walk_to functions
As @vonagam suggested in #68 .

Add an `offset: Vector2` parameter to `walk_to_clicked` , `walk_to_prop` , `walk_to_hotspot` , `walk_to_marker` in **PopochiuCharacter**. It will make sense when moving multiple characters since "it makes sense to place them not at the same exact position".

vonagam:
Closes #69.

Also changed type for `E.clicked` from generic `Node` to `PopochiuClickable`, since `hovered` already has that type and only clickable changes that variable that should not be a problem I assume.


mapedorr (approved):
Everything looks good. I tested the changes and they work pretty well.

PR #73, opened 2023-07-30
Finish transition from point to marker

vonagam:
There is unfinished migration in naming from Point to Marker in 2.0. Some of it just names, but some does create errors - like calls to non-existent method `get_point` on a room (it is called now `get_marker`).

- Rename `Constants.Types.POINT` to `Constants.Types.MARKER`.
- In Room interface remove `get_point`, add `get_marker` and `get_marker_position`.
- In Character class make `walk_to_marker` consistent with other walk_to methods - if marker does not exist, do not move.
- In Room class make `get_marker` return `Marker2D` (you get what you ask, plus `get_markers` return array of those too, marker can provide scale or rotation, so there is a value in it), add `get_marker_position` to match old `get_point` behavior.

Alternatively, `get_marker_position` can be named `get_point` to match old name. Thoughts?

### DIFF
--- a/addons/popochiu/editor/main_dock/tab_room.gd
+++ b/addons/popochiu/editor/main_dock/tab_room.gd
@@ -39,7 +39,7 @@ var _remove_dialog: ConfirmationDialog
 		type_class = PopochiuRegion,
 		parent = 'Regions'
 	},
-	Constants.Types.POINT: {
+	Constants.Types.MARKER: {
 		group = find_child('MarkersGroup'),
 		method = 'get_markers',
 		type_class = Marker2D,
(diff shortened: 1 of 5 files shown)


mapedorr:
We had a discussion about the `get_marker` vs `get_point` in Discord, and we decided that, now that Godot 4 uses **Marker2D**, renaming everything to _marker_ was the right option. Furthermore, the `get_point` (and the Point concept itself) was something we initially copied from PowerQuest to maintain consistency between the two plugins.

So, I'll go with the changes you're proposing. But, before doing so, I want to make sure it will not generate too much conflict with the changes I'm reviewing in #64 .

Thanks for your help and patience!

Issue #71, opened 2023-07-25
Add a toolbar button and popup panel to handle in-game audio.


mapedorr:
Add a button to the toolbar to open a popup panel to handle the volume of the different audio buses in the game.

![audio_toolbar](https://github.com/mapedorr/popochiu/assets/4536477/0e0164b2-4863-4c8d-a4b7-d6c0d87bf846)

Each channel could be a button, so clicking it will restore the volume of that channel to its default value.

![audio_popup](https://github.com/mapedorr/popochiu/assets/4536477/ed002413-216b-4531-ae65-d4f6da98476b)

For creating the popup, you can use the Save/Load scene (`res://addons/popochiu/engine/objects/graphic_interface/save_load/save_load.tscn`) as reference.

> About the popup, this is just a suggestion. We can discuss a different layout and functionality for it.

PR #74, opened 2023-08-04
refs #71: Toolbar audio settings

JuannFerrari:
This PR is for Issue #71 

**What**
This PR adds a new audio settings panel with sliders and saves them into a file, if the file exists the saved settings will be loaded.

**Why**
To let the player modify the audio from the game.

**Before merge:**
Set a proper icon for the toolbar
image

**Screenshot**
image


stickgrinder:
@JuannFerrari I've left a couple very minor comments. Apart from that, this is very good work to me! Thanks! :)

addons/popochiu/engine/objects/graphic_interface/sound_settings/sound_settings.gd
```diff
@@ -0,0 +1,61 @@
...
+	show()
+
+func _update_master_volume(value: float) -> void:
+	am.set_bus_volume_db("Master", value)
+	
+func _update_music_volume(value: float) -> void:
+	am.set_bus_volume_db("Music", value)
+
+func _update_sfx_volume(value: float) -> void:
+	am.set_bus_volume_db("Effects", value)
```
mapedorr, line 54: Instead of having three methods, we should have only one that changes the volume of the corresponding bus based on the parameter it receives.

addons/popochiu/engine/objects/graphic_interface/sound_settings/sound_settings.tscn
```diff
@@ -0,0 +1,100 @@
...
+unique_name_in_owner = true
+layout_mode = 2
+min_value = -80.0
+max_value = 0.0
+
+[node name="SFXSlider" type="HSlider" parent="SoundSettingsDialog/VBoxContainer/HBoxContainer/VBoxContainer"]
+unique_name_in_owner = true
+layout_mode = 2
+min_value = -65.0
+max_value = 0.0
```
mapedorr, line 88: These should be created during runtime (except for Master) if we decide to leave developers to define their own buses.

mapedorr (changes requested):
Apart from these comments, before we proceed with the merge, we'll need to determine the icon for the button on the toolbar. Let me know if you'd like me to take care of that and provide it to you.

addons/popochiu/engine/objects/graphic_interface/toolbar/toolbar.tscn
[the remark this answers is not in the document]
mapedorr: Yes. We could think about this now that I'm working on refactoring/improving the GUI.

mapedorr (changes requested):
While testing the Audio settings I found a couple of bugs that we need to fix before merging the PR:

- [x] If you create a new audio bus and then run the game, Godot shows an error in the Output. You can still open the Audio settings popup and see the name of the new bus as part of the buses list, but it is not possible to change its volume.
- [x] Supose you have four buses: Master, Music, SFX, and Voices. If you delete the SFX bus, the Voices bus will show/edit the volume of the SFX bus instead of updating its own volume.
- [x] After installing the plugin on an empty project, Godot shows an error when running the game because it cannot open the `default_bus_layout.tres` file.
- [x] Once you create the `default_bus_layout.tres` file and run the project, Godot shows an error saying that the `volume_settings` variable in **audio_manager.gd** doesn't find an entry for `'Master'`.
- [x] If one deletes the `user://audio_settings.save` file, running the game shows the same error as the previous bug.

mapedorr (approved):
Very well done!

Issue #101, opened 2023-10-24
Add anti-glide animation feature


Whyshchuck:
It would be nice to add an option for characters to run his animation in anti-glade mode like.

How it work - if anti-glade is on chaa=racter's sprite cheneges its position only when new walking cycles' frame is displayed.
It avoids sprite movement between frames - avoids effect of gliding.

It would work like AGS method ```Character.MovementLinkedToAnimation```
https://adventuregamestudio.github.io/ags-manual/Character.html?highlight=glide&case_sensitive=0#charactermovementlinkedtoanimation

Why it is important and when it is necessary:
If you have sharp eye, you would find it necessary in every game, where:
- walking cycle FPS is lower than game's FPS and
- distance traveled by character beetween frames is bigger than 1 pixel.

In effect it will be most important in low res games with low fps animations

Below short animation illustrating anti-glide in action (walking cycle is not perfect but I've slowed down animation and characters movement to make it more illustrous)

https://github.com/mapedorr/popochiu/assets/34949896/85dc9385-e873-4b49-a97d-167e756be09a

mapedorr:
I think this can be a must have for the plugin. And as you already made some progress on the code necessary to make it work, we could think about adding it to v2 release.

PR #102, opened 2023-10-25
refs 101: anti-glide animation added

Whyshchuck:
Following changes adds option for anti-glide movement.

In order to make it working, for every character:
- add call method track in walking cycle animations (for every direction) that will call ```position_update()``` every frame 
image
- chenge anti-glide animation field to "on":
image

for further description check #101

### DIFF
--- a/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
+++ b/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
@@ -293,16 +293,15 @@ func _move_along_path(distance):
 			#track in animation player and call update_position every frame
 			#additional condition to avoid crash if update_position method
 			#would be deleted
+			var next_position = last_point.linear_interpolate(
+					_path[0], distance / distance_between_points
+				)
 			if (_moving_character.anti_glide_animation == true 
 			and _moving_character.has_method("update_position")
 			):
-				moving_character_position_stored = last_point.linear_interpolate(
-					_path[0], distance / distance_between_points
-				)
+				moving_character_position_stored = next_position
 			else:
-				_moving_character.position = last_point.linear_interpolate(
-					_path[0], distance / distance_between_points
-				)
+				_moving_character.position = next_position
 			return
 
 		distance -= distance_between_points
--- a/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
+++ b/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
@@ -282,10 +282,7 @@ func _move_along_path(distance):
 			# Play the animation on the character
 			_moving_character.play_walk(_path[0])
 			# Move the character on the destination
-			#in order to run anti-glide animation you need to add call method 
-			#track in animation player and call update_position every frame
-			#additional condition to avoid crash if update_position method
-			#would be deleted
+			
 			var next_position = last_point.linear_interpolate(
 					_path[0], distance / distance_between_points
 				)
--- a/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
+++ b/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
@@ -21,6 +21,13 @@ var _path := []
 var _moving_character: PopochiuCharacter = null
 var _nav_path: PopochiuWalkableArea = null
 
+#stores character position between frames if anti-glide animation is on
+var moving_character_position_stored = (
+	_moving_character.position 
+	if _moving_character 
+	else null
+	)
+
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ GODOT ░░░░
 func _enter_tree() -> void:
@@ -259,10 +266,20 @@ func clean_characters() -> void:
 		if c is PopochiuCharacter:
 			c.queue_free()
 
+func update_characters_position(character):
+	character.position = (
+		moving_character_position_stored 
+		if moving_character_position_stored 
+		else character.position
+		)
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ PRIVATE ░░░░
 func _move_along_path(distance):
-	var last_point = _moving_character.position
+	var last_point = (
+		moving_character_position_stored 
+		if moving_character_position_stored 
+		else _moving_character.position
+		)
 	
 	while _path.size():
 		var distance_between_points = last_point.distance_to(_path[0])
@@ -272,9 +289,20 @@ func _move_along_path(distance):
 			# Play the animation on the character
 			_moving_character.play_walk(_path[0])
 			# Move the character on the destination
-			_moving_character.position = last_point.linear_interpolate(
-				_path[0], distance / distance_between_points
-			)
+			#in order to run anti-glide animation you need to add call method 
+			#track in animation player and call update_position every frame
+			#additional condition to avoid crash if update_position method
+			#would be deleted
+			if (_moving_character.anti_glide_animation == true 
+			and _moving_character.has_method("update_position")
+			):
+				moving_character_position_stored = last_point.linear_interpolate(
+					_path[0], distance / distance_between_points
+				)
+			else:
+				_moving_character.position = last_point.linear_interpolate(
+					_path[0], distance / distance_between_points
+				)
 			return
 
 		distance -= distance_between_points
--- a/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
+++ b/addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
@@ -21,14 +21,6 @@ var _path := []
 var _moving_character: PopochiuCharacter = null
 var _nav_path: PopochiuWalkableArea = null
 
-#stores character position between frames if anti-glide animation is on
-var moving_character_position_stored = (
-	_moving_character.position 
-	if _moving_character 
-	else null
-	)
-
-
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ GODOT ░░░░
 func _enter_tree() -> void:
 	if Engine.editor_hint:
@@ -257,7 +249,7 @@ func get_characters() -> Array:
 	for c in $Characters.get_children():
 		if c is PopochiuCharacter:
 			characters.append(c)
-	
+			
 	return characters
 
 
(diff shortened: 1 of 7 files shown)


Whyshchuck:
things to expolore:
- maybe side-scrolling could be updated to base on ```moving_character_position_stored``` instead of ```position```
-  should be tested with multiple characters moving - maybe ```moving_character_position_stored``` should be characters' field or array to address such cases

addons/Popochiu/Engine/Templates/CharacterTemplate.gd
```diff
@@ -5,6 +5,9 @@ const Data := preload('CharacterStateTemplate.gd')
 
 var state: Data = null
 
+# turns anti-glide animation on.
+# will not work if call method track in animation player is not calling update_position every frame
+export var anti_glide_animation: bool = false
```
mapedorr, line 10: Must be in `PopochiuCharacter.gd`, which is the class from where all characters inherit from.

This script is only the one used as the template for developers to write their logic.

addons/Popochiu/Engine/Templates/CharacterTemplate.gd
```diff
@@ -52,3 +55,9 @@ func play_talk() -> void:
 # Use it to play the grab animation for the character
 func play_grab() -> void:
 	.play_grab()
+
+# Updates character position in the current room if anti-glide animation is on
+# should be called every frame in call method track in animation player
+func update_position():
+	E.current_room.update_characters_position(C.PLAYER_2)
```
mapedorr, line 62: Must be in `PopochiuCharacter.gd`, which is the class from where all characters inherit from. And it should send `self` instead of `C.PLAYER_2`.

This script is only the one used as the template for developers to write their logic.

popochiu/Characters/Goddiu/CharacterGoddiu.gd
```diff
@@ -8,6 +8,9 @@ const Data := preload('CharacterGoddiuState.gd')
 
 var state: Data = preload('CharacterGoddiu.tres')
 
+# turns anti-glide animation on.
+# will not work if call method track in animation player is not calling update_position every frame
+export var anti_glide_animation: bool = false
```
mapedorr, line 13: Not needed if you move the variable to `PopochiuCharacter.gd` as mentioned in the comment for the `CharacterTemplate.gd` file.

This comment also applies for `CharacterPopsy.gd` and `CharacterTrapusinsiu.gd`.

popochiu/Characters/Goddiu/CharacterGoddiu.gd
```diff
@@ -48,3 +51,9 @@ func play_talk() -> void:
 # Use it to play the grab animation for the character
 func play_grab() -> void:
 	.play_grab()
+
+# Updates character position in the current room if anti-glide animation is on
+# should be called every frame in call method track in animation player
+func update_position():
+	E.current_room.update_characters_position(self)
```
mapedorr, line 58: Not needed if you move the function to `PopochiuCharacter.gd` as mentioned in the comment for the `CharacterTemplate.gd` file.

This comment also applies for `CharacterPopsy.gd` and `CharacterTrapusinsiu.gd`.

mapedorr (changes requested):
Looking good. Just needs a small refactor.

addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
```diff
@@ -272,9 +289,20 @@ func _move_along_path(distance):
...
+			if (_moving_character.anti_glide_animation == true 
+			and _moving_character.has_method("update_position")
+			):
+				moving_character_position_stored = last_point.linear_interpolate(
+					_path[0], distance / distance_between_points
+				)
+			else:
+				_moving_character.position = last_point.linear_interpolate(
+					_path[0], distance / distance_between_points
+				)
```
edmundito, line 305: Since it's the same math, it could be done before the if statement to simplify the statement
```suggestion
			var next_position := last_point.linear_interpolate(
					_path[0], distance / distance_between_points
			)
			if (_moving_character.anti_glide_animation
			  and _moving_character.has_method("update_position")
			):
				moving_character_position_stored = next_position
			else:
				_moving_character.position = next_position
```

edmundito (commented):
Just a minor comment. Also agreed with @mapedorr 's review.

addons/Popochiu/Engine/Objects/Room/PopochiuRoom.gd
edmundito: Looks like the indentation on my suggestion is weird, I would suggest not applying this change directly but using it as a guideline.

Whyshchuck: @edmundito @mapedorr thanks for your comments. I will apply them together with changing moving_character_position_stored  into making PopochiuCharacter field (to avoid interference between multiple moving characters)

addons/Popochiu/Engine/Templates/CharacterTemplate.gd
Whyshchuck: oh, I've already  fixed it but in Goddiu not in template...

Whyshchuck:
And for the full happiness we need one more thing.
Some script, that would edit character's .scn file and from this:
```tracks/4/type = "value"
tracks/4/path = NodePath("Sprite:frame")
tracks/4/interp = 1
tracks/4/loop_wrap = false
tracks/4/imported = false
tracks/4/enabled = true
tracks/4/keys = {
"times": PoolRealArray( 0, 0.1 ),
"transitions": PoolRealArray( 1, 1 ),
"update": 1,
"values": [ 4.0, 5.0 ]
}```
would make this:
```tracks/5/type = "method"
tracks/5/path = NodePath(".")
tracks/5/interp = 1
tracks/5/loop_wrap = true
tracks/5/imported = false
tracks/5/enabled = true
tracks/5/keys = {
"times": PoolRealArray( 0, 0.1 ),
"transitions": PoolRealArray( 1, 1 ),
"values": [ {
"args": [  ],
"method": "update_position"
}, {
"args": [  ],
"method": "update_position"
} ]
}```
it would make it easy to implement without manual adding method calls
@stickgrinder maybe you would have an idea how to do it the best way?

**Review by mapedorr — APPROVED** [review]

**Review by mapedorr — APPROVED** [review]
Great job! I'm gonna make a couple of tests before merging this to main.

**Comment by stickgrinder** [comment]
> Or maybe we could loop over "times": PoolRealArray in order to call method without adding additional track Everything to emulate work of frame_changed ( ) signal from animated sprite class

OK, me @mapedorr just took a whole hour to com up with a possible cleaner solution, just to realize that may be a very simpler one... The `AnimationPlayer` emits no signal when the frame is changed because it makes no sense, you may not be animating the frame property of a Sprite2D, but the Sprite2D itself emits a signal when its frame changes.

So we have two ways to know if the frame changed without the need to expose a function on the character:

1) The room just listen to the `Character.Sprite2D.frame_changed" signal of the moving character and if the anti glide is on AND the character is moving, the position gets updated in your "quantized" way.

2) We can hack things a bit so that _our_ animation player actually emits a signal, but for that we should create a child class:

```
class_name PopochiuAnimationPlayer
extends AnimationPlayer

signal animation_frame_changed

func _post_process_key_value(animation, track, value, object, object_idx):
	if animation.track_get_path(track).get_subname(0) in ["frame"]:
		animation_frame_changed.emit(animation)
	
	# This is necessary for the animation player to actually continue process the animation
	return value
 ```
 
 and change the type in the base PopochiuCharacter scene to match this one.
This virtual is invoked at every keyframe of every track. Here we check that we are processing a keyframe on the right target and if so, we know the frame did change.

I guess going with option 1 is way better, but just to report our findings, this second method seems to work.

Do you feel like restructuring the calling chain so that it all belongs to PopochiuRoom?

Whyshchuck:
> > Or maybe we could loop over "times": PoolRealArray in order to call method without adding additional track Everything to emulate work of frame_changed ( ) signal from animated sprite class
> 
> OK, me @mapedorr just took a whole hour to come up with a possible cleaner solution, just to realize that may be a very simpler one... The `AnimationPlayer` emits no signal when the frame is changed because it makes no sense, you may not be animating the frame property of a Sprite2D, but the Sprite2D itself emits a signal when its frame changes.
> 
> So we have two ways to know if the frame changed without the need to expose a function on the character:
> 
> 1. The room just listen to the `Character.Sprite2D.frame_changed" signal of the moving character and if the anti glide is on AND the character is moving, the position gets updated in your "quantized" way.
> 2. We can hack things a bit so that _our_ animation player actually emits a signal, but for that we should create a child class:
> 
> ```
> class_name PopochiuAnimationPlayer
> extends AnimationPlayer
> 
> signal animation_frame_changed
> 
> func _post_process_key_value(animation, track, value, object, object_idx):
> 	if animation.track_get_path(track).get_subname(0) in ["frame"]:
> 		animation_frame_changed.emit(animation)
> 	
> 	# This is necessary for the animation player to actually continue process the animation
> 	return value
> ```
> 
> and change the type in the base PopochiuCharacter scene to match this one. This virtual is invoked at every keyframe of every track. Here we check that we are processing a keyframe on the right target and if so, we know the frame did change.
> 
> I guess going with option 1 is way better, but just to report our findings, this second method seems to work.
> 
> Do you feel like restructuring the calling chain so that it all belongs to PopochiuRoom?

cool solutions. My first approach to anti-glide was based on AnimatedSprite and its signal. Whole idea of call method track came out of the lacking signal.

One question: how both slutions will work in case of:
- change of name from eg. Sprite2D to CharacterSpriteWalkingCycles
- multiple Sprite nodes?
Can we target this-particular-sprite-that-is-displayed-at-the-moment?

stickgrinder:
> One question: how both slutions will work in case of:
> 
>     * change of name from eg. Sprite2D to CharacterSpriteWalkingCycles
> 
>     * multiple Sprite nodes?
>       Can we target this-particular-sprite-that-is-displayed-at-the-moment?

Are those just speculations or do you have specific use cases in mind? The same sprite is used for every animation ATM since the whole character animations are exported as a single spritesheet.

We have options anyway: the important part is to be able to target the nodes we want to connect to, so we may group them or use the node metadata to label them somehow, for example.

One basic principle towards which Popochiu is moving is providing tools to achieve the most common sets of use-cases, and don't get in the way if game devs want/need to leave the paved road. But if this happens they maybe have to build their own very specific tools.

Ideally the most mature version of Popochiu we have in mind will keep devs very far from the scene tree :)

If you can explain a bit more of what you have in mind, I can think of a possible solution.
Thanks!

Whyshchuck:
> If you can explain a bit more of what you have in mind, I can think of a possible solution. Thanks!

Not everyone will necesarily use the importer. If so, there is no problem to add and parse new sprites textures for newly added animations instead of updating already existing

stickgrinder:
> Not everyone will necesarily use the importer. If so, there is no problem to add and parse new sprites textures for newly added animations instead of updating already existing

Sorry if I gave the impression I was implying anything about the importer, it was not the case.

I was just assuming exporting a single spritesheet was the preferred option anyway, but of course I get why someone would arrange things differently.

In that case, I would exclude solution 2 and go with 1, the problem "just" being how to instruct the room about which Sprite2Ds to track.

This can be solved programmatically or by exposing an array of sort in the inspector maybe? I admit I have no clear ideas on how to organize that second case.

If we do it programmatically, should the room subscribe to all Sprite2D available in characters, this would not be a problem I guess, because if the character is not walking we can just ignore signals. If it's walking, we are pretty sure the engine is playing a walk animation of sort, no matter which Sprite2D is actually being updated.

I don't like too much automagic in these cases but it seems pretty safe... Am I missing something?

Whyshchuck:
@stickgrinder  i have connected all children sprites with update_position function. Now it works without animation call method track

stickgrinder:
After a quick chat with @Whyshchuck we decided that this contribution, very relevant and already available in 2.0, won't have much impact on 1.x version, that is about to be tagged as unmaintained legacy.

Closing this PR.
Thanks anyway

Issue #87, opened 2023-10-06
Audio Cues: Checking "loop" renders sound silent


papernoise:
**Godot**: 4.1.2
**Popochiu**: 2.0 (latest version from 2.0 branch)

## How to reproduce:

- Add a sound to Popochiu (eg. mx_music.wav)
- Click the play icon: no sound can be heard
- Trigger the sound via code: no sound can be heard

Unticking "loop" will play the sound just fine.
My guess is that the loop is set to  zero samples or something like that

mapedorr:
@JuannFerrari Added a warning in Godot's Output panel with #104 .

PR #104, opened 2023-11-09
#87 Add warning when loop metadata is missing for AudioStreamWAV

JuannFerrari:
This PR is for issue #87 

**What**
This PR adds a warning in the Godot console to help developers notice that the WAV file added does not have the correct metadata to loop

**Why**
Is not a bug from Popochiu per se, according to https://github.com/godotengine/godot/issues/40359 if the metadata is not present a WAV file won't loop.

### DIFF
--- a/addons/popochiu/engine/audio_manager/audio_cue.gd
+++ b/addons/popochiu/engine/audio_manager/audio_cue.gd
@@ -93,9 +93,12 @@ func set_loop(value: bool) -> void:
 		'AudioStreamOggVorbis', 'AudioStreamMP3':
 			audio.loop = value
 		'AudioStreamWAV':
-			(audio as AudioStreamWAV).loop_mode =\
-			AudioStreamWAV.LOOP_FORWARD if value\
-			else AudioStreamWAV.LOOP_DISABLED
+			if (audio as AudioStreamWAV).get_loop_end() == 0 && value:
+				push_warning("res://addons/popochiu/engine/audio_manager/audio_cue.gd:97 " + resource_name + " does not have the correct metadata for loop, please check AudioStreamWAV documentation")
+			else:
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/audio_manager/audio_cue.gd
```diff
@@ -93,9 +93,12 @@ func set_loop(value: bool) -> void:
 		'AudioStreamOggVorbis', 'AudioStreamMP3':
 			audio.loop = value
 		'AudioStreamWAV':
-			(audio as AudioStreamWAV).loop_mode =\
-			AudioStreamWAV.LOOP_FORWARD if value\
-			else AudioStreamWAV.LOOP_DISABLED
+			if (audio as AudioStreamWAV).get_loop_end() == 0 && value:
+				push_warning("res://addons/popochiu/engine/audio_manager/audio_cue.gd:97 " + resource_name + " does not have the correct metadata for loop, please check AudioStreamWAV documentation")
```
mapedorr, line 97: I suggest using this way of displaying the warning so that it is understood as something triggered by the plugin and avoid long code lines:

```gdscript
PopochiuUtils.print_warning(
	"[b]%s[/b]" % resource_name +\
	" does not have the correct metadata to loop, please check" +\
	" AudioStreamWAV documentation"
)
```

mapedorr (commented):
Just a suggestion to keep messages consistency throught the whole plugin.

addons/popochiu/engine/audio_manager/audio_cue.gd
JuannFerrari: that's great, I will change it

mapedorr (approved):
Thanks!

PR #125, opened 2023-12-13
refs 101: anti-glide animation addedanti-glide animation added (for v2.0)

Whyshchuck:
This is variant of https://github.com/mapedorr/popochiu/pull/102 for Popochiu 2.0
Supports movement of multiple characters at once.

tu switch between anti-glide and normal movement - use checkbox in character inspector (separately for every character)
image

for further description check #101


mapedorr (approved):
Really nice job.

Code is clean and works perfectly.

Issue #114, opened 2023-12-07
Character sprite scales depending on its position in rooms


stickgrinder:
### Benefit description

That's a pretty common feature in adventure games, not worth a detailed description. Basically, if a room has a deep perspective, it's necessary to scale the character as it walks toward the back of the scene, far from the camera.

### Solution description

Character scaling may depend on the single walking area. A scene may have more than one (think Thimbleweed Park Cemetery, where the character travels over a path between small hills), and each one can be positioned at a certain distance from the camera.

Each WA should expose two values representing the scaling fraction for the bottom-most and top-most points in the WA. The engine can then interpolate based on the vertical distance between those points.

### Exclusions

Upscaling is known to provide poor results, so we should probably exclude the possibility of an upscaled character (open discussion).

### Implications

* If the scale factor is exposed as a fraction or percentage (between 1 and 0), 0 should be an allowed value.
* If more than a single walk areas are available in the same room, should consistency between them be owned by the developer? This is the best option to allow the support for strange perspective that may be very useful in large "fisheye" or panorama views (think Phatt Island dock in Monkey Island 2)

stickgrinder:
This was left open by mistake, @Whyshchuck already provided this feature.

PR #127, opened 2023-12-13
refs 114: character sprite scaling regions

Whyshchuck:
Every region in the room can be set as "scaling":

image

than you can set scale for the top and bottom of the region (meant as the highest and lowest position of region's polygon

every character stores info about scaling region he's standing on, in form of dictionary.
than game calculates scale proportionate (linear) for the character's opsition.

Warnings:
- needs additional changes to be merged with my anti-glide (i have it ready, but i can't pull request it in reasonable way until it's merged).
- acts a bit wirdeley in one edge case:
    if there is drastic change of scale at the top of the region, character enters and exits region multiple times.
    to avoid it, collishion shape for character's have to be reduced to [0,0, 0,0, 0,0, 0,0]

For further infos #114

### DIFF
--- a/addons/popochiu/engine/objects/region/popochiu_region.gd
+++ b/addons/popochiu/engine/objects/region/popochiu_region.gd
@@ -41,8 +41,15 @@ func _update_scaling_region(chr: PopochiuCharacter) -> void:
 		'region_description': self.description,
 		'scale_top': self.scale_top, 
 		'scale_bottom': self.scale_bottom,  
-		'polygon_top_y': polygon_y_array.min()+self.position.y+get_node("InteractionPolygon").position.y if self.position else '',
-		'polygon_bottom_y': polygon_y_array.max()+self.position.y+get_node("InteractionPolygon").position.y if self.position else '',
+		'polygon_top_y': (
+			polygon_y_array.min()+self.position.y+get_node("InteractionPolygon").position.y 
+			if self.position 
+			else ''
+			),
+		'polygon_bottom_y': (
+			polygon_y_array.max()+self.position.y+get_node("InteractionPolygon").position.y 
+			if self.position 
+			else ''),
 
 		}
 
--- a/addons/popochiu/engine/objects/region/popochiu_region.gd
+++ b/addons/popochiu/engine/objects/region/popochiu_region.gd
@@ -11,10 +11,9 @@ extends Area2D
 # TODO: If walkable is false, characters should not be able to walk through this.
 #export var walkable := true
 @export var tint := Color.WHITE
-# TODO: Make the scale of the character change depending checked where it is placed in
-# the area.
-#export var scale_top := 1.0
-#export var scale_bottom := 1.0
+@export var scaling :bool = false
+@export var scale_top :float = 1.0
+@export var scale_bottom :float = 1.0
 
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ GODOT ░░░░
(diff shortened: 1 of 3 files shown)


addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -33,12 +32,35 @@ func _on_character_entered(chr: PopochiuCharacter) -> void:
...
+	var polygon_y_array = []
+	for x in get_node("InteractionPolygon").get_polygon():
+		polygon_y_array.append(x.y)
+
+	chr.on_scaling_region= {
+		'region_description': self.description,
+		'scale_top': self.scale_top, 
+		'scale_bottom': self.scale_bottom,  
+		'polygon_top_y': polygon_y_array.min()+self.position.y+get_node("InteractionPolygon").position.y if self.position else '',
+		'polygon_bottom_y': polygon_y_array.max()+self.position.y+get_node("InteractionPolygon").position.y if self.position else '',
```
mapedorr, line 45: To prevent lines of code exceeding column 100, I suggest performing the operations to calculate `polygon_top_y` and `polygon_bottom_y` like this.

```suggestion
		'polygon_top_y': (
			polygon_y_array.min()+self.position.y+get_node("InteractionPolygon").position.y
		) if self.position else '',
		'polygon_bottom_y': (
			polygon_y_array.max()+self.position.y+get_node("InteractionPolygon").position.y
		) if self.position else '',
```

addons/popochiu/engine/objects/room/popochiu_room.gd
```diff
@@ -287,6 +287,26 @@ func set_active_walkable_area(walkable_area_name: String) -> void:
...
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ PRIVATE ░░░░
+
+func _update_character_scale(chr):
+	if chr.on_scaling_region:
+		var polygon_range = chr.on_scaling_region['polygon_bottom_y'] - chr.on_scaling_region['polygon_top_y']
+		var scale_range = chr.on_scaling_region['scale_bottom'] - chr.on_scaling_region['scale_top']
+
+		var position_from_the_top_of_region = chr.position.y-chr.on_scaling_region['polygon_top_y']
+
+		var scale_for_position = (
```
mapedorr, line 298: - Same as previous comment. I would prefer to avoid having lines of code that go beyond line 100.
- When possible, define the type of the variables you're using.

```suggestion
		var polygon_range: float = (
			chr.on_scaling_region['polygon_bottom_y'] - chr.on_scaling_region['polygon_top_y']
		)
		var scale_range: float = (
			chr.on_scaling_region['scale_bottom'] - chr.on_scaling_region['scale_top']
		)

		var position_from_the_top_of_region: float = (
			chr.position.y-chr.on_scaling_region['polygon_top_y']
		)

		var scale_for_position: float = (
```

mapedorr (changes requested):
The code works perfectly. I just left a couple of suggestions related with variable types and the length of some code lines.

mapedorr:
There is another edge case (or is it a bug?):

- If you create a region that encloses the character in its original position, scaling won't be applied until it starts moving. (sorry for the audio in the video)

https://github.com/mapedorr/popochiu/assets/4536477/7a30221f-9770-46cc-89dc-3ab24d6775ad

Whyshchuck:
> There is another edge case (or is it a bug?):
> 
> * If you create a region that encloses the character in its original position, scaling won't be applied until it starts moving. (sorry for the audio in the video)
> 
>  no_scaling_until_moves.mp4

that's right. I've checked it onli with anti-glide and then _update_position was called fust after sprite load.
Without anti_glide - no. Now it should be fixed!

mapedorr (approved):
Ready to merge!

Thanks for taking the comments into account.

PR #141, opened 2023-12-24
Refs 114: character sprite scaling regions update

Whyshchuck:
This is update to my earlier PR.
It resolves 2 bugs observed during further tests:
- characters with bigger collision polygons had tendency to scale exceeding maximum and minimum scale set for top and bottom of region;
-  in some cases scale didn't applied to characters before room's fade in. They either shrinked just after load or didnt scale until any character moved.
Both issues should now be fixed
- scaling works only for main room. Regions in all other rooms have scaling=False and scale_top and _bottom ==1

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -26,6 +26,9 @@ var last_room := ''
 var anim_suffix := ''
 var is_moving := false
 var emotion := ''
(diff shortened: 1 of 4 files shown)


mapedorr (changes requested):
I just ask for a change on those lines of code that are too long in `popochiu_room.gd`.

mapedorr (approved):
Ready to merge!!!

PR #142, opened 2023-12-28
Ref 114: Character sprite scaling update - scaling polygon added

Whyshchuck:
characters' InteractionPolygon was a bad candidae to rule  entering and exiting of scaling regions.
- InteractionPolygons needed to be as large as sprite to allow clickability of character;
- polygons used to determine moment of entering or exiting of scaling regions needed to be as small as possible in order to avoid sequence of enters and exits while character shrinks at the region's border.

That made resonable creation of separate Polygon for scaling purpose - ie. ScalingPolygon.

### DIFF
--- a/addons/popochiu/engine/objects/region/popochiu_region.gd
+++ b/addons/popochiu/engine/objects/region/popochiu_region.gd
@@ -22,6 +22,8 @@ func _ready() -> void:
 	
 	area_entered.connect(_check_area.bind(true))
 	area_exited.connect(_check_area.bind(false))
+	area_shape_entered.connect(_check_scaling.bind(true))
+	area_shape_exited.connect(_check_scaling.bind(false))
 
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ VIRTUAL ░░░░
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -61,17 +63,23 @@ func _clear_scaling_region(chr: PopochiuCharacter) -> void:
 		chr.on_scaling_region = {}
 
 # ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ PRIVATE ░░░░
+func _check_scaling(area_rid: RID, area: Area2D, area_shape_index: int, local_shape_index: int, entered: bool):
+	if area is PopochiuCharacter and area.get_node("ScalingPolygon") and area_shape_index == area.get_node("ScalingPolygon").get_index():
```
mapedorr, line 67: I'll just ask you to break the function parameters declaration and the first condition so they don't exceed column 100.

Also, sometimes Godot doesn't do well when comparing `get_node(...)`, so I prefer to use `get_node_or_null(...)`.

```suggestion
func _check_scaling(
	area_rid: RID, area: Area2D, area_shape_index: int, local_shape_index: int, entered: bool
):
	if (
		area is PopochiuCharacter and area.get_node_or_null("ScalingPolygon")
		and area_shape_index == $ScalingPolygon.get_index()
	):
```

mapedorr (changes requested):
Just a minor change related with long lines of code.

mapedorr (approved):

PR #143, opened 2023-12-29
Refs 114: Character sprite scaling fix

Whyshchuck:
In order to solve problems listed in 
https://github.com/mapedorr/popochiu/pull/141
room's method update_character_scale function name had been changed as it's private no more.
unfortunately in next PR:
https://github.com/mapedorr/popochiu/pull/142
one of references in popochiu_region restored the outdated name.

Following PR covers fix of: 
- above mentioned bug and
- avoids transfer of characters' position to the next room (via character's position_stored variable)


stickgrinder:
LGTM, merging

PR #146, opened 2024-01-05
Refs 145: walking on broken line fix

Whyshchuck:
Following solution fixes problem of character playing once oriented animation from  the very beginning up to the end of animation, ignoring geometry of broken line path.

In order to fix that ```_play_walk``` had to be called in order to have ```_get_valid_oriented_animation called```.

Reservations:
```_play_walk``` is private function. In order to make this solution clean, it should be made public or should be wrapped into some public function eg.:

```
func take_turn(target_pos: Vector2) -> void:
    _play_walk(target_pos)
```


stickgrinder:
@Whyshchuck LGTM, thanks a lot! :+1: 

@mapedorr I leave this to you to merge so you can rebase when you prefer

PR #148, opened 2024-01-08
Refs 145: walking on broken line encapsulated

Whyshchuck:
Following change updates https://github.com/carenalgas/popochiu/pull/146 in order to follow encapsulation rules.

```_play_walk``` and ```face_direction``` had been wrapped into ```take_turn``` public function which can be freely called buy the room script.


mapedorr:
I understand the purpose of the function, but I'm curious about its name. Why `take_turn`?

Whyshchuck:
> I understand the purpose of the function, but I'm curious about its name. Why `take_turn`?

Because it's called every time character takes turn while walking

Issue #147, opened 2024-01-05
More robust project documentation


stickgrinder:
### Benefit description

Popochiu outgrew the incubation stage and now requires better, more mature documentation that:

- Provides a user experience on par with Godot itself so that users will feel at home
- Can be hosted on GitHub Pages (or ReadTheDocs if we decide to use it in the future; that's where Godot's docs belong)
- It's sourced from the same repository as the project itself so that we can enforce documentation writing in our _Definition of Done_ for new features or breaking changes
- It's based on MarkDown, so we can make use of the work that's already been done
- Can be partially generated automatically from code comments (see #133) without breaking manual contributions
- Supports switch for different engine releases/versions
- Can be deployed to a subdirectory or subdomain in GHPages in case we want to add a Popochiu website at root level in the future

From a content point of view:

- is LOOSELY based on the concepts of the [Diataxis framework](https://diataxis.fr/), to the larger possible extent
- can be properly organized and themed depending on the project's needs

### Solution description

We scouted for a solution and decided to go with [MkDocs](https://www.mkdocs.org/) for it's natively supported by RTD and can be easily deployed on GHPages.

Also, compared to Sphinx, it uses MarkDown, and its configuration is a single YAML file.

The todo-list follows:

* [x] Kickstart the docs project in the main repo
* [x] Activate automation to deploy
* [x] ~~Bring the currently available documentation to the repo~~ (split in #185)
* [x] Provides instructions or presets for contributors to run MkDocs locally, if needed (docker-based solution is preferred even in Windows, but we should consider not littering the repo. Maybe a sub-repo?)
* [x] ~~Clean up the wiki or close it entirely~~ (split in #185)
* [x] ~~Review the project's Readme to redirect to the GHPages homepage~~ (split in #186)

### Exclusions

We're not going to create a custom theme.  We'll customize the logo and colors, and that's all.

### Implications

This will allow contributors to add content without access to the project's wiki repo, which is way more problematic.

stickgrinder:
I'm closing this because I split some tasks in more granular issues.

The bulk of the docs framework is done, it's now time to move on with the content.

PR #156, opened 2024-02-17
refs #147: Code docblocks exporter

stickgrinder:
This PR introduces a first implementation of a codeblock exporter which makes it possible to "dump" all the inline Godot documentation for Engine classes to markdown files, so everything that gets documented in code, can also be navigated in the site.

## Featured in the PR

* A customized version of [gdscript-docs-maker](https://github.com/GDQuest/gdscript-docs-maker) that address issues with invalid JSON file exported from Godot when numeral constants (like `INF` or `NAN`) are used as params default values.
* A better, custom docker container to run _gdscript-docs-maker_ (only 266MB instead of > 1.5GB, plus Godot and GDM versions are pinned to the working ones)
* Make targets to build and run the GDM container on any OS that supports Docker Desktop and Make (instructions provided in a dedicated README)
* A syntax highlighter for GDScript based on `highlight.js,` already integrated and custom-themed for our documentation :)

## Not featured, but planned

* GitHub Action to export the refs before publishing the docs to production
* Resorting to a published and maintained docker image by the Carenalgas community, downloadable from DockerHub
* A way to better sort the reference. At the moment, the navigation is pretty awkwardly ordered
* More extensive doc blocks, with keywords to structure information in MarkDown
* Contributions to upstream GDM repo, that will eventually be done

And, of course... COMPLETE TESTING!


mapedorr (approved):
I tested it in WSL and it worked!

Awesome work.

Issue #170, opened 2024-03-06
Hover text and Cursor are wrong after interaction


mapedorr:
### Bug description

When a hotspot or prop or character is clicked, the cursor keeps showing its hover text even though it comes out of its InteractionPolygon. Same for the cursor, which keeps showing the "active" animation (in the 2-click Context-sensitive GUI).

Thanks to **_antecanis78_** for reporting the bug in Discord and providing us the video.

###  Steps to reproduce

Using the 2-click Context-sensitive GUI.

1. Create a room with a hotspot and make the player say something when it is clicked.
2. Play the game and move the mouse over the hotspot. The HoverText will show its description and the cursor will play the "active" animation.
3. While the character is saying its dialogue line, move the cursor out of the hotspot.
4. The HoverText will keep showing the description of the hotspot, and the cursor will still play the "active" animation, even though the cursor is not hovering the object.

### Expected vs observed behavior

The HoverText should change if the cursor leaves the hotspot. The cursor should go back to its "normal" state too.

https://github.com/carenalgas/popochiu/assets/4536477/4d8f1870-8c59-4064-accc-45b8f92bac25

### Environment information (please complete):

 - **OS type and version**: Windows 11
 - **Godot version**: 4.2.1
 - **Plugin version or branch**: Beta 1

stickgrinder:
Fixed in #172

Issue #166, opened 2024-03-03
Can't block interactions


mapedorr:
### Bug description

When `PopochiuCharacter.walk_to_clicked` is called, there's no way to block interactions until the action finishes, which causes undesired behaviors.

###  Steps to reproduce

1. Create a Hotspot and make the character walk to it when clicked and then say something.
2. Run the game and click the created Hotspot.
3. While the character is moving towards the Hotspot, click elsewhere.
4. The character will move to the last clicked position, and after that will say the dialogue line scripted for the Hotspot.

### Expected vs observed behavior

There should be a way to block interactions when `PopochiuCharacter.walk_to_clicked` is called. Maybe the default behavior should block interactions, or maybe there should be an alternate method like `PopochiuCharacter.walk_to_clicked_blocking`.

### Environment information (please complete):

 - **OS type and version**: Windows 11
 - **Godot version**: 4.2.1
 - **Plugin version or branch**: beta 1

stickgrinder:
Closed in #172

PR #172, opened 2024-03-15
Fix for 170 and 166

mapedorr:
For #170 : **PopochiuClickable** no longer stops listening mouse entering/exiting while the GUI is blocked. And the HoverText and the cursor are restored to default when moving to another room.

For #166 : After the movement ends, the character's position is checked. If the position doesn't match the expected one, execution is indefinitely paused (`await E.await_stopped`). Conversely, if the position matches, the `script_name` of the object that triggered the movement is compared to `E.clicked.script_name` to verify if the currently clicked object is the same as the one that triggered the action.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -497,7 +497,13 @@ func queue_walk_to_clicked(offset := Vector2.ZERO) -> Callable:
 ## Makes the character walk to the last clicked [PopochiuClickable], which is stored in
 ## [member Popochiu.clicked]. You can set an [param offset] relative to the target position.
 func walk_to_clicked(offset := Vector2.ZERO) -> void:
+	var clicked_id: String = E.clicked.script_name
+	
 	await _walk_to_node(E.clicked, offset)
+	
+	# A double check in case two PopochiuClickable share the same `walk_to_point` coordinates
+	if clicked_id != E.clicked.script_name:
+		await E.await_stopped
 
 
 func walk_to_clicked_blocking(offset := Vector2.ZERO) -> void:
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -246,6 +246,7 @@ func walk(target_pos: Vector2) -> void:
 	# Trigger the signal for the room to start moving the character
 	started_walk_to.emit(self, position, target_pos)
 	await move_ended
+	
 	is_moving = false
 
 func take_turn(target_pos: Vector2):
@@ -419,12 +420,6 @@ func say(dialog: String, emo := "") -> void:
 		await get_tree().process_frame
 		return
 	
-	# Blocks the graphic interface so players can't interact with it while the
-	# dialog line plays
-	# NOTE: What if players want NPCs talking without blocking the graphic
-	# 		interface?
-	G.block()
-	
 	if not emo.is_empty():
 		emotion = emo
 	
@@ -441,10 +436,6 @@ func say(dialog: String, emo := "") -> void:
 	
 	emotion = ''
 	idle()
-	
-	# Unblock the graphic interface with a delay to prevent cursor flickering
-	# (rapid state changes) between multiple lines of text or actions in sequence
-	G.unblock(true)
 
 
 ## Calls [method _play_grab] and waits until the [signal grab_done] is emitted, then goes back to
@@ -507,6 +498,14 @@ func queue_walk_to_clicked(offset := Vector2.ZERO) -> Callable:
 ## [member Popochiu.clicked]. You can set an [param offset] relative to the target position.
 func walk_to_clicked(offset := Vector2.ZERO) -> void:
 	await _walk_to_node(E.clicked, offset)
+
+
+func walk_to_clicked_blocking(offset := Vector2.ZERO) -> void:
+	G.block()
+	
+	await _walk_to_node(E.clicked, offset, true)
+	
+	G.unblock()
 
 
 ## Makes the character walk to the [PopochiuProp] (in the current room) which
@@ -811,14 +810,21 @@ func _get_valid_oriented_animation(animation_label):
 	return null
 
 
-func _walk_to_node(node: Node2D, offset: Vector2) -> void:
+func _walk_to_node(node: Node2D, offset: Vector2, is_blocking := false) -> void:
 	if not is_instance_valid(node):
 		await get_tree().process_frame
 		return
-
-	await walk(
+	
+	var target_pos := (
 		node.to_global(node.walk_to_point if node is PopochiuClickable else Vector2.ZERO) + offset
 	)
+	var target_move_pos: Vector2 = R.current.get_move_target_position(position, target_pos)
+	
+	await walk(target_pos)
+	
+	if not is_blocking and position != target_move_pos:
+		await E.await_stopped
+
 
 func _update_position():
 	E.current_room.update_characters_position(self)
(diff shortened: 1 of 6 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -811,14 +810,21 @@ func _get_valid_oriented_animation(animation_label):
 	return null
 
 
-func _walk_to_node(node: Node2D, offset: Vector2) -> void:
+func _walk_to_node(node: Node2D, offset: Vector2, is_blocking := false) -> void:
```
stickgrinder, line 819: I don't want to become overly pedantic, but boolean params like these should be avoided because they make the calls to the function unreadable. What does `true` stand for?

It may be better to have constants like `ACT_MODE_BLOCKING`, `ACT_MODE_NON_BLOCKING`, so that:

1. It's always clear what you are attempting
2. You can push those down to public methods, if the need be, because of point 1
3. Should you stumble onto another necessary mode of execution, it's easier to add a constant; with bools you're stuck with a two-way switch

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -811,14 +810,21 @@ func _get_valid_oriented_animation(animation_label):
...
-	await walk(
+	
+	var target_pos := (
 		node.to_global(node.walk_to_point if node is PopochiuClickable else Vector2.ZERO) + offset
 	)
+	var target_move_pos: Vector2 = R.current.get_move_target_position(position, target_pos)
+	
+	await walk(target_pos)
+	
+	if not is_blocking and position != target_move_pos:
```
stickgrinder, line 831: I'm probably misreading this, but shouldn't it be:

"If IS BLOCKING and you are not yet at destination, then await the movement is over" ?

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -95,8 +95,8 @@ func _ready():
 		
 		# Connect to singleton signals
 		E.language_changed.connect(_translate)
-		G.blocked.connect(_on_graphic_interface_blocked)
-		G.unblocked.connect(_on_graphic_interface_unblocked)
+		#G.blocked.connect(_on_graphic_interface_blocked)
```
stickgrinder, line 98: If you think those lines are useless now, I'd clean those up.

addons/popochiu/engine/popochiu.gd
```diff
@@ -1023,7 +1024,8 @@ func _on_character_spoke(chr: PopochiuCharacter, msg := '') -> void:
 
 
 func _on_graphic_interface_unblocked() -> void:
-	clicked = null
+	pass
```
stickgrinder, line 1027: Duh?!

Can you please explain?

addons/popochiu/engine/objects/room/popochiu_room.gd
```diff
@@ -405,6 +405,12 @@ func set_active_walkable_area(walkable_area_name: String) -> void:
 		PopochiuUtils.print_error("Can't set %s as active walkable area" % walkable_area_name)
 
 
+func get_move_target_position(start_position: Vector2, end_position: Vector2) -> Vector2:
```
stickgrinder, line 408: Is this a public utility function? Does it need to be documented (API wise?)

stickgrinder (changes requested):
Left some comments. Sorry if I misread something.

One is worth a change, in my personal opinion. Let me know.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -497,7 +497,13 @@ func queue_walk_to_clicked(offset := Vector2.ZERO) -> Callable:
 ## Makes the character walk to the last clicked [PopochiuClickable], which is stored in
 ## [member Popochiu.clicked]. You can set an [param offset] relative to the target position.
 func walk_to_clicked(offset := Vector2.ZERO) -> void:
+	var clicked_id: String = E.clicked.script_name
+	
 	await _walk_to_node(E.clicked, offset)
+	
+	# A double check in case two PopochiuClickable share the same `walk_to_point` coordinates
+	if clicked_id != E.clicked.script_name:
```
stickgrinder, line 505: Sorry man, but I don't get what are you doing here. I see nothing wrong, so I'm just curious. Can you please explain?

addons/popochiu/engine/objects/character/popochiu_character.gd
mapedorr: Nope.

The purpose of this `if` statement is to verify that the final position of the movement is as expected only when the action is NON-BLOCKING. Otherwise, there is no need to perform the validation because it is assumed that the character will reach its destination since the GUI will be blocked, preventing clicks from being registered.

addons/popochiu/engine/objects/character/popochiu_character.gd
mapedorr: I get your point. Right now I can not imagine another mode of execution appart from BLOCKING and NON-BLOCKING, but I agree that a new one could appear in the future. I'm gonna make the change because I like the idea of making things more readable.

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
mapedorr: Yes. I always forget to remove commented lines. Thanks!

addons/popochiu/engine/popochiu.gd
mapedorr: My fault. I forgot to remove the function.

addons/popochiu/engine/objects/room/popochiu_room.gd
mapedorr: Yes. I forgot to document it. Thanks!

addons/popochiu/engine/objects/character/popochiu_character.gd
mapedorr: Imagine a scenario where two props/hotspots/characters have equal values in their `walk_to_point` properties. Since the cancellation of movement depends on the character's final position matching the expected target position (`popochiu_character.gd:825`), the code execution would continue because the condition `if not is_blocking and position != target_move_pos` evaluates to `false`. To prevent this from happening, I added an additional check: that the object that initiated the character's movement is the same active object (`E.clicked`) once the movement has finished. I know that by doing this, we could simplify everything to just this single check (ignoring the position comparison). I just kept both because I wasn't sure... or because I'm an idiot.

If you think it's clearer to just leave this latter one, I could get rid of several things that raised doubts in other comments.

mapedorr: Actually... I think I will leave only this one.

mapedorr: Furthermore, there might be another option that would avoid having to compare the values of `script_name`: having a boolean property in the **PopochiuClickable** class (e.g., `execution_cancelled`) that is set to `true` when an interaction interrupts the execution of instructions (e.g., clicking on the scene or another object).

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: Ah OK, got it.
Thanks

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: I saw your last commit, and I think this is way cleaner.

To me this is mergeable!

stickgrinder (approved):

PR #174, opened 2024-03-17
refs #147: Docs are deployed with every new release + Release should now work again

stickgrinder:
This PR completes the Popochiu 2.0 documentation framework, adding a GitHub action that starts alongside the "New release" one.

I also added a first commit with a permission fix that should bring back the release workflow (it was broken).

I tested:

- That the deploy via GH Actions works as intended. It does.
- That we can still use the `make docs-deploy` from any branch and it does not clash with the GH Actions WF. Everything seems ok.

I didn't test:

- If the release WF is actually working again. We have to push a `v2.0-beta2` tag for that. But it should be ok now since I clearly identified the root cause.


mapedorr (approved):
We'll test this with the release of beta 2.

Issue #188, opened 2024-03-25
Scale GUI not working


mapedorr:
### Bug description

GUI is not scaling even though the `scale_gui` property in the Settings is set to `true`.

###  Steps to reproduce

1. Create a project with a Native game resolution greater than the default one (i.e. 1280x720 (HD)).
2. Check the Scale GUI property in Settings (it should be checked by default).
3. Run the game with any GUI template activated.
4. The GUI appears at its original size instead of scaling based on the scale value shown in the Setup popup (for a 1280x720 game resolution the scale is 4.0).

![image](https://github.com/carenalgas/popochiu/assets/4536477/092b577e-5065-4468-bd09-5bd4f585f234)

### Expected vs observed behavior

The GUI should scale to match the value mentioned in the Setup popup.

### Environment information (please complete):

 - **OS type and version**: Windows 11
 - **Godot version**: 4.2.1
 - **Plugin version or branch**: Beta 2

### Additional context

This is how the 9 Verb GUI looks like:

![image](https://github.com/carenalgas/popochiu/assets/4536477/f4543959-b1d9-48c4-8478-9b0434874fb2)

And this is how the Sierra GUI looks like:

![image](https://github.com/carenalgas/popochiu/assets/4536477/ca2e7451-9775-4837-9cbf-40ec970ef1b9)

Additionally, when opening the settings menus of these two interfaces (or the inventory in the Sierra template), the popups are not visible.

PR #193, opened 2024-04-02
GUI Scaling (fix for #188)

mapedorr:
Makes `PopochiuSettings.scale_gui` to be taken into account by the plugin to scale the GUI. It also makes this scaling work on the 3 templates available so far.

[Here](https://mega.nz/folder/frQzFQqD#lWMaQNnBP0l6Dm-yu5ymng) are the some assets you can use to create the scene to test creating a game in HD (1280x720).

## How the scaling works (open to discussion):

1. If scaling is enabled, Popochiu sets the size of the GUI (the root node in the `res://game/graphic_interface/graphic_interface.tscn`) to its original size (320x180) **during runtime**, and then scales that node to the value calculated in the Setup window.
2. Some GUI components take this scale value into account to change the way elements are rendered:
    - **InventoryBar** and **InventoryGrid** make the texture of items to expand (using `TextureRect.EXPAND_FIT_WIDTH`) so they fit the height of the inventory slot. I made this to assure devs don't have to design their assets (in this case the PNGs of each inventory item) in a smaller size. E.j. if the image of each inventory item is 64x64, it will be scaled up during runtime because the whole GUI is scaled up.
    - **Cursor** scales the texture of selected inventory items (this affects the Sierra and 2-click Context-sensitive GUIs) to match the height of the texture used in by the main cursor (this is the cursor used by default, when no inventory item is selected). Similar to the previous point, this ensures that the texture used by the secondary cursor (the one used to render selected inventory items) doesn't look bigger than the one used by the main cursor.

## Known issues with this approach

- Devs will have to know that the images for the inventory items will be scaled down by Popochiu during runtime.
- It is hard to know the right amount of space occupied by the GUI when editing the `.tscn` files inside `res://game/graphic_interface`. Since the scaling is done during runtime, this is how the GUI looks in the Editor:
  
  image

  ☝️ **This is the one that concerns me the most at the moment** 🙈 .

- Godot has some issues rendering fonts. So, not all of them will look good. Most are readable, but others look weird. Here's an example of how the Settings window in the 9 Verb GUI is rendered in a Full HD (1280x720) game (look how "Resume Game" is rendered):

  image

## Other approaches I've been thinking of

- More than scaling, is having a different theme for each GUI template in which the font and the assets are designed for HD (or Full HD) games and beyond.
- Instead of scaling the GUI at runtime, we could scale the components (I'm referring to the root nodes inside the .tscn files) that are copied to the `res://game/graphic_interface/` folder. This way, developers will see that the scenes are scaled in the Editor itself.

### DIFF
--- a/addons/popochiu/engine/cursor/cursor.gd
+++ b/addons/popochiu/engine/cursor/cursor.gd
@@ -86,8 +86,8 @@ func set_secondary_cursor_texture(texture: Texture2D, ignore_block := false) -> 
 	secondary_cursor.texture = texture
 	
 	if E.settings.scale_gui:
-		# Scale the cursor based on the size of the relation of the texture size against
-		# the main cursor size
+		# Scale the cursor based the relation of the texture size compared to the main cursor
+		# texture size
 		secondary_cursor.scale = Vector2.ONE * ceil(
 			float(texture.get_height()) / float(get_cursor_height())
 		)
--- a/addons/popochiu/engine/cursor/cursor.gd
+++ b/addons/popochiu/engine/cursor/cursor.gd
@@ -21,6 +21,9 @@ enum Type {
 
 var is_blocked := false
 
+@onready var main_cursor: AnimatedSprite2D = $MainCursor
+@onready var secondary_cursor: Sprite2D = $SecondaryCursor
+
 
 #region Godot ######################################################################################
 func _ready():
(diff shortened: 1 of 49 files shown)


addons/popochiu/engine/cursor/cursor.gd
```diff
@@ -67,35 +70,43 @@ func show_cursor(anim_name := "normal", ignore_block := false) -> void:
 	
 	if (
 		not anim_name.is_empty()
-		and not $MainCursor.sprite_frames.has_animation(anim_name)
+		and not main_cursor.sprite_frames.has_animation(anim_name)
 	):
-		prints("[Popochiu] Cursor has no animation: %s" % anim_name)
+		PopochiuUtils.print_error("Cursor has no animation: %s" % anim_name)
```
stickgrinder, line 75: Nice one @mapedorr, can you please explain this util function so that I can take note of that and document it for contributors?

addons/popochiu/engine/cursor/cursor.gd
```diff
@@ -67,35 +70,43 @@ func show_cursor(anim_name := "normal", ignore_block := false) -> void:
...
 func set_secondary_cursor_texture(texture: Texture2D, ignore_block := false) -> void:
 	if not ignore_block and is_blocked: return
 	
-	$SecondaryCursor.texture = texture
+	secondary_cursor.texture = texture
+	
+	if E.settings.scale_gui:
+		# Scale the cursor based the relation of the texture size compared to the main cursor
+		# texture size
+		secondary_cursor.scale = Vector2.ONE * ceil(
```
stickgrinder, line 91: Great use of `Vector2.ONE`, I am an idiot with these things.

addons/popochiu/engine/cursor/cursor.gd
mapedorr: Sure.

I created a couple of functions to print messages in the output with a special style (using bbcode) so we can use when a message is from our plugin. At the moment, we have: `print_error(msg: String)` which shows the message in red, `print_warning(msg: String)` which shows the message in yellow, and `print_normal(msg: String)` which shows the message in green. All of them use the colors of the palette used by Popochiu: https://lospec.com/palette-list/colodore.

Now that you mentioned, I have to finish the code documentation for the **PopochiuUtils** class.

Issue #92, opened 2023-10-07
Double clicking on characters or objects activates it twice


BenjaTK:
This gets pretty annoying if, for example, you talk to a character and they double their dialogue. Especially if you have the `walk_to_clicked` animation, you can still interact with props when that's happening.

mapedorr:
Fixed in #198 thanks to @anthonyirwin82 .

PR #198, opened 2024-04-06
Added double click support

anthonyirwin82:
I wanted to be able to use a double click function on hotspots to be able to have an exit hotspot instantly change rooms instead of having to wait for the character to walk there.

Added _on_double_click() to work the same as _on_click() and _on_right_click functions do.

### DIFF
--- a/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
+++ b/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
@@ -236,6 +236,7 @@ func on_click() -> void:
 
 ## Called when the object is double clicked.
 func on_double_click() -> void:
+	PopochiuUtils.reset_double_click()
 	_on_double_click()
 
 
@@ -257,7 +258,7 @@ func on_item_used(item: PopochiuInventoryItem) -> void:
 ## Triggers the proper GUI command for the clicked mouse button identified with [param button_idx],
 ## which can be [enum MouseButton].MOUSE_BUTTON_LEFT, [enum MouseButton].MOUSE_BUTTON_RIGHT or
 ## [enum MouseButton].MOUSE_BUTTON_MIDDLE.
-func handle_command(button_idx: int, double_click: bool = false) -> void:
+func handle_command(button_idx: int) -> void:
 	var command: String = E.get_current_command_name().to_snake_case()
 	var prefix := "on_%s"
 	var suffix := "click"
@@ -266,10 +267,7 @@ func handle_command(button_idx: int, double_click: bool = false) -> void:
 		MOUSE_BUTTON_RIGHT:
 			suffix = "right_" + suffix
 		MOUSE_BUTTON_MIDDLE:
-			suffix = "middle_" + suffix
-	
-	if button_idx == MOUSE_BUTTON_LEFT and double_click:
-		suffix = "double_" + suffix
+			suffix = "middle_" + suffix	
 
 	if not command.is_empty():
 		var command_method := suffix.replace("click", command)
@@ -339,10 +337,18 @@ func _on_mouse_exited() -> void:
 
 
 func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int):
-	if not PopochiuUtils.is_click_or_touch_pressed(event): return
 	if not E.hovered or E.hovered != self: return
 	
-	var event_index := PopochiuUtils.get_click_or_touch_index(event)
+	if await PopochiuUtils.is_double_click_or_tap(event):
+		times_double_clicked += 1
+		E.clicked = self
+		on_double_click()
+		return
+	
+	if not await PopochiuUtils.is_click_or_touch_pressed(event): return
+			
+	
+	var event_index := await PopochiuUtils.get_click_or_touch_index(event)
 	E.clicked = self
 	last_click_button = event_index
 	
@@ -353,12 +359,8 @@ func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int):
 			if I.active:
 				on_item_used(I.active)
 			else:
-				if event.double_click:
-					handle_command(event_index, true)
-					times_double_clicked += 1
-				else:
-					handle_command(event_index)
-					times_clicked += 1
+				handle_command(event_index)
+				times_clicked += 1
 		MOUSE_BUTTON_RIGHT, MOUSE_BUTTON_MIDDLE:
 			if I.active: return
 			
--- a/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
+++ b/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
@@ -52,6 +52,11 @@ var last_click_button := -1 # NOTE Don't know if this will make sense, or if thi
 
 @onready var _description_code := description
 
+## Used for setting the double click delay. Windows default is 500 miliseconds.
+static var double_click_delay: float = 0.2 # 0.5 felt like too long of a delay before acting
+## Used for tracking if a double click has occured.
+static var has_double_click: bool = false
+
 
 #region Godot ######################################################################################
 func _ready():
@@ -236,7 +241,7 @@ func on_click() -> void:
 
 ## Called when the object is double clicked.
 func on_double_click() -> void:
-	PopochiuUtils.reset_double_click()
+	reset_double_click()
 	_on_double_click()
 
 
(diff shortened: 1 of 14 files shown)


addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -244,7 +257,7 @@ func on_item_used(item: PopochiuInventoryItem) -> void:
 ## Triggers the proper GUI command for the clicked mouse button identified with [param button_idx],
 ## which can be [enum MouseButton].MOUSE_BUTTON_LEFT, [enum MouseButton].MOUSE_BUTTON_RIGHT or
 ## [enum MouseButton].MOUSE_BUTTON_MIDDLE.
-func handle_command(button_idx: int) -> void:
+func handle_command(button_idx: int, double_click: bool = false) -> void:
```
mapedorr, line 260: I suggest not changing the signature of this method since it's likely that the double click will only be considered when it's a left click (or a double tap on a touch screen). I'll explain myself in detail in the comment for line 357..

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -255,6 +268,9 @@ func handle_command(button_idx: int) -> void:
 		MOUSE_BUTTON_MIDDLE:
 			suffix = "middle_" + suffix
 	
+	if button_idx == MOUSE_BUTTON_LEFT and double_click:
+		suffix = "double_" + suffix
+
```
mapedorr, line 273: This won't be necessary if we keep the signature of the method intact.

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -337,17 +353,20 @@ func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int):
 			if I.active:
 				on_item_used(I.active)
 			else:
-				handle_command(event_index)
-				
-				times_clicked += 1
+				if event.double_click:
+					handle_command(event_index, true)
```
mapedorr, line 357: It might not be necessary to call `handle_command()` here. If only clicks (or taps) can be double, we can directly call `on_double_click()` here.

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -337,17 +353,20 @@ func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int):
 			if I.active:
 				on_item_used(I.active)
 			else:
-				handle_command(event_index)
-				
-				times_clicked += 1
+				if event.double_click:
```
mapedorr, line 356: To accommodate touchscreen devices, it would be best to create a method in **PopochiuUtils** that checks for double clicks or double taps. I suggest something like this:
```
static func is_double_click_or_tap(event: InputEvent) -> bool:
    if is_click_or_touch_pressed(event):
        if event is InputEventMouseButton:
            return event.double_click
        elif event is InputEventScreenTouch:
            return event.double_tap
    return false
```

addons/popochiu/engine/templates/hotspot_template.gd
```diff
@@ -17,6 +17,14 @@ func _on_click() -> void:
...
 
 
+func _on_double_click() -> void:
+	# Replace the call to E.command_fallback() with your code.
+	E.command_fallback()
+	# For example, you could make the player instantly goto another room on an exit hotspot instead
+	# of waiting for the player to walk there.
+#	await E.goto_room("NewRoom")
+
+
```
mapedorr, line 27: The `_on_double_click()` virtual method should be included in the templates for props and characters too.

mapedorr (changes requested):
Just take this into account (I'm not asking you to fix it, since I've been doing some tests and the solution doesn't looks trivial):

The `_on_click()` is triggered before `_on_double_click()`, so if you have logic for the `_on_click()` and excecute a double click, both actions will be triggered. It might be better to introduce a small delay to ensure that the event to be triggered is indeed the double click. This is beneficial because we can use it to address issue #92 .

If you want to try something for this, go ahead. Otherwise, take the comments I did into account and I'll merge the PR once they are done. I can take care of fixing the triggering of both events.

addons/popochiu/engine/others/popochiu_utils.gd
```diff
@@ -3,6 +3,11 @@ class_name PopochiuUtils
 extends Node
 ## Utility functions for Popochiu.
 
+## Used for setting the double click delay. Windows default is 500 miliseconds.
+static var double_click_delay: float = 0.35 # 0.5 felt like too long of a delay before acting
+## Used for tracking if a double click has occured.
+static var has_double_click: bool = false
```
mapedorr, line 9: Just to mention, in this case, because the types are `float` and `bool` there is no need to specify those in the variable declaration. You can just simply do: `static var double_click_delay := 0.35` and Godot will know it is a float. Same for the `bool` variable.

mapedorr:
After making a couple of tests I'm gonna say I don't like the delay after clicking on things. I would like to see if we can reduce it to 0.1 or even lower. Additionally, with this approach, skipping the animation of the dialogue lines for characters when clicking anywhere on the screen stopped working. That causes the character to move while saying a dialogue line (and I don't like the idea of forcing players to use cutscenes).

Taking these changes as a starting point, I'd like to experiment a bit tomorrow to see if I can reach a solution. If I find that it's going to require many changes, I think the most appropriate course of action will be to remove the wait (and all the `await`s related with the click/tap), merge the PR, and then spend more time solving the problem. Something tells me that the root cause lies in a recent change I made in how clicks are detected in **PopochiuClickable**s.

addons/popochiu/engine/others/popochiu_utils.gd
anthonyirwin82: My understanding is if you used the types on the variables then it improves the auto code completion and can also help with godot giving you warnings or errors on your code before running the game.

The documentation also suggests fully using typed or dynamic code for consistency.

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html

If you want to have functions typed and variable declarations not typed then that is easy enough to do. I am wondering how that effects code completion in particular. I have not used godot enough to know the difference in code completion quality particularly as I tend to add the types on the little things I have done so far.

anthonyirwin82:
When I was looking at default millisecond delay and found that Windows was 500 milliseconds and tested that it was painful, that's why I lowered it. I tested what you said about making it 0.1 and found that I sometimes had a single click and double click event come through.

When I tested 0.2 as a value, I did not get any extra unwanted click events and did not notice any delay in movement.

Another thing to think about is we are going to need a click and hold event to have the verb coin interface where you click and hold on a hotspot then a verb coin pops up with the different interaction types.

People making touch screen games may also what a two finger tap and hold to show all the game hotspots. I have also seen single tap and hold on the player character show all the hotspots.

Other events people may want is to have wheel mouse up and down to show or hide the inventory bar.

This is all stuff that visionaire studio can do so is probably something to add at some point.

I have not used popochiu enough to know how the dialog text worked prior to me making changes for the double click. I am pretty sure other engines disable mouse clicking when dialog is playing unless they use a say_background() or something which allows normal gameplay to continue while text is being displayed. This allows a character locked in a jail cell to say something random like "please get me out of here", or a loudspeaker announcing things but does not stop the player from moving.

mapedorr:
I tested the update, and 0.2 seconds feel better. We can keep that delay by now while we check if there is a way to avoid it.

We don't disable the mouse while the dialog lines are showing, the GUI just captures the click and stops its propagation. But with the delay that stopped working.

We'll have a `say_background()` in future versions because other devs have been asking for it. Meanwhile, and before merging your PR, I want to address the issue with the dialog lines. Perhaps after releasing version 2, we can add a configuration option where you can specify whether dialogue lines should or should not block the character's movement, but for now, I would like it to remain as it has worked so far. Here's an example of the desired behavior:

https://github.com/carenalgas/popochiu/assets/4536477/6f52ec0c-51f7-4978-8049-3188b1c84c05

Another issue I found is that, if you move the cursor after clicking anywhere in the room, the target direction changes if one moves the cursor before the character starts walking:

https://github.com/carenalgas/popochiu/assets/4536477/541cf323-20e4-40be-ba78-db94b8f80fe0

addons/popochiu/engine/others/popochiu_utils.gd
anthonyirwin82: I see what you mean.

var f1 = 1.0 # dynamic
var f2 := 1.0 # statically typed with inference

mapedorr (approved):

PR #200, opened 2024-04-13
Move Popochiu settings config to ProjectSettings (fix #26)

### LINKED ISSUE #26: Popochiu settings are moved to Project Settings
At the moment, the plugin settings are exposed in the inspector as a set of resource properties, because it was quicker than creating a dedicated interface.

immagine

Having those exposed in the Project Settings would be sexier though (and sexier is better than quickier).

mapedorr:
This moves the settings related to Popochiu to a section in the ProjectSettings.

### DIFF
--- a/addons/popochiu/editor/config/config.gd
+++ b/addons/popochiu/editor/config/config.gd
@@ -1,4 +1,5 @@
 @tool
+class_name PopochiuConfig
 extends RefCounted
 
 enum DialogStyle {
@@ -9,12 +10,6 @@ enum DialogStyle {
 	BUBBLE_ABOVE_CHARACTER,
 }
 
-# EDITOR SETTINGS ----------------------------------------------------------------------------------
-const ASEPRITE_IMPORTER_ENABLED = "popochiu/aseprite_import/enable_aseprite_importer"
-const ASEPRITE_COMMAND_PATH = "popochiu/aseprite_import/command_path"
-const ASEPRITE_REMOVE_JSON_FILE = "popochiu/aseprite_import/remove_json_file"
-
-# PROJECT SETTINGS ---------------------------------------------------------------------------------
 # Thanks to @drbloop for providing the bases of the new approach for moving the popochiu settings to
 # Godot's ProjectSettings instead of using a Resource file.
 # ---- GUI -----------------------------------------------------------------------------------------
@@ -46,21 +41,9 @@ const ASEPRITE_WIPE_OLD_ANIMATIONS = "popochiu/aseprite_import/wipe_old_animatio
 const PIXEL_ART_TEXTURES = "popochiu/pixel/pixel_art_textures"
 const PIXEL_PERFECT = "popochiu/pixel/pixel_perfect"
 
-var editor_settings: EditorSettings
-
-var _plugin_icons: Dictionary
-
 
 #region Public #####################################################################################
-func initialize_editor_settings():
-	editor_settings = EditorInterface.get_editor_settings()
-	
-	_initialize_editor_cfg(ASEPRITE_IMPORTER_ENABLED, false, TYPE_BOOL)
-	_initialize_editor_cfg(ASEPRITE_COMMAND_PATH, _default_command(), TYPE_STRING)
-	_initialize_editor_cfg(ASEPRITE_REMOVE_JSON_FILE, true, TYPE_BOOL)
-
-
-func initialize_project_settings():
+static func initialize_project_settings():
 	# ---- GUI -------------------------------------------------------------------------------------
 	_initialize_project_cfg(SCALE_GUI, false, TYPE_BOOL)
 	# TODO: Move this to the properties of the 2-click Context-sensitive template or its InventoryBar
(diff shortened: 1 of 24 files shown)


addons/popochiu/editor/config/config.gd
```diff
@@ -1,127 +1,210 @@
...
-func _get_project_setting(key: String, default_value):
+
+static func is_pixel_perfect() -> bool:
+	return _get_project_setting(PIXEL_PERFECT, false)
+
+
+#endregion
+
+#region Private ####################################################################################
+static func _initialize_project_cfg(
```
stickgrinder, line 184: It's just nitpicking, but what if this is called `_initialize_project_setting()` instead? I know it come in the `config.gd` file so maybe it doesn't make sense, but it would be more readable about what it does in the new context.

addons/popochiu/editor/importers/aseprite/animation_creator.gd
```diff
@@ -185,7 +182,7 @@ func _load_spritesheet_metadata(selected_tag: String = ""):
 	_spritesheet_metadata.sprite_sheet = sprite_sheet
 
 	# Remove the JSON file if config says so
-	if _config.should_remove_source_files():
+	if PopochiuEditorConfig.should_remove_source_files():
```
stickgrinder, line 185: Aaaaawwwwwwww!!! :star_struck:

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -594,6 +594,9 @@ func play_animation(animation_label: String, animation_fallback := 'idle'):
 			[script_name]
 		)
 		return
+	
+	if $AnimationPlayer.get_animation_list().is_empty():
```
stickgrinder, line 598: This smells like a bugfix sneaked into this PR, right?

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -820,8 +823,7 @@ func _get_valid_oriented_animation(animation_label):
 		var animation = "%s%s" % [animation_label, suffix]
 		if $AnimationPlayer.has_animation(animation):
 			return animation
-	# No valid animation is found.
-	printerr('Animation not found %s' % [animation_label])
```
stickgrinder, line 824: This was probably intended as useful information for the developer, but I agree an error is too much.
Do you think we can log this info somewhere so that the dev has a log of missing animations that the engine expects to be played?

addons/popochiu/engine/objects/popochiu_settings.gd
```diff
@@ -3,62 +3,46 @@ class_name PopochiuSettings
...
-# 		a cleaner way to give devs tools to customize this. Maybe we can take a look to the
-# 		SceneManager (https://github.com/glass-brick/Scene-Manager) plugin to see how it handles
-# 		transitions.
-## A reference to the scene used as to handle transition animations between rooms and other game
-## events.
-## This will be [color=bf5a50]deprecated[/color].
-@export var transition_layer: PackedScene = null
 ## The time, in seconds, that will take the game to skip a cutscene.
-@export var skip_cutscene_time := 0.2
+var skip_cutscene_time := 0.0
```
stickgrinder, line 7: Why the value change?

stickgrinder: Forget me, it's clear.

addons/popochiu/engine/popochiu.gd
```diff
@@ -152,23 +152,16 @@ func _ready() -> void:
 	am = load(PopochiuResources.AUDIO_MANAGER).instantiate()
 	
 	# Set the Graphic Interface node
-	if settings.graphic_interface:
-		gi = settings.graphic_interface.instantiate()
-		gi.name = 'GraphicInterface'
-	else:
-		gi = load(PopochiuResources.GUI_ADDON_FOLDER).instantiate()
+	gi = load(PopochiuResources.GUI_GAME_SCENE).instantiate()
```
stickgrinder, line 155: Is this ALWAYS available due to the default? What if a game has a custom GUI selected and no GUI is available to instantiate?

addons/popochiu/popochiu_plugin.gd
```diff
@@ -53,36 +51,33 @@ func _enter_tree() -> void:
 	# Good morning, starshine. The Earth says hello.
 	prints(ES)
 	prints(EN)
-	print_rich('[wave]%s[/wave]' % SYMBOL)
+	print_rich("[wave]%s[/wave]" % SYMBOL)
```
stickgrinder, line 54: At this point I'm going to ask: should we always prefer double over single quotes for strings in the source code?

addons/popochiu/editor/config/config.gd
```diff
@@ -1,127 +1,210 @@
...
+	ProjectSettings.save()
 
-var _plugin_icons: Dictionary
 
+# ---- GUI -----------------------------------------------------------------------------------------
+static func is_scale_gui() -> bool:
+	return _get_project_setting(SCALE_GUI, false)
 
-#region Public #####################################################################################
-func initialize_editor_settings():
```
stickgrinder, line 24: Not sure where this one ended up. Are we still going to have editor settings for the plugin?

addons/popochiu/editor/config/config.gd
mapedorr: I agree. I'll do the change.

addons/popochiu/engine/objects/character/popochiu_character.gd
mapedorr: More than a bugfix, I see it as a small UX improvement. Many messages like this appear in the Output:

image

They can be annoying during the early stages of development if animations are not yet ready, or if one has characters without animations.

This makes sure the logic about triggering the right animation is not done if there are no animations in the character.

addons/popochiu/engine/objects/character/popochiu_character.gd
mapedorr: I agree, but we already have a message that gives the proper context to devs. You can see an example of it in the screenshot that is part of the answer to your previous comment. I think there is no need to print this message if we already have a better one: it uses the Popochiu print style and gives a better feedback.

addons/popochiu/engine/popochiu.gd
mapedorr: What do you mean by "a custom GUI selected"?

The GUI scene should be always available because Popochiu creates the corresponding files for the GUI when one closes the Setup popup the first time. But thanks to your comment, I just remembered that I have to include a validation to prompt the user to select a GUI in case they come from a version of the plugin prior to beta 1 (I don't remember in which version we included the GUI templates).

addons/popochiu/popochiu_plugin.gd
mapedorr: I prefer double because it makes it easier to create messages that have apostrophes. Do you prefer single quotes?

addons/popochiu/editor/config/config.gd
mapedorr: Yes, but now they are in the `res://addons/popochiu/editor/config/editor_config.gd` script.

PR #203, opened 2024-04-18
Fixed the code to correctly create the popochiu-skip action automatic…

anthonyirwin82:
When enabling the popochiu addon and reloading the project the popochiu-skip action for skipping cutscenes was not created automatically.

This has been fix in this pull request.


mapedorr (approved):
Nice!

PR #206, opened 2024-04-19
added support for character to teleport to markers, hotspots and props

anthonyirwin82:
Other adventure game engines I have used provided helper functions to both walk places with walking animations or to instantly move from their current location to a new location without any animations.

I have added teleport helper functions the same as the existing walk_to ones for teleporting the character from their current location to a new one. This should make it easier for new users as they don't need to learn how to do it with gdscript and provides similar functionality to other adventure game engines.

queue_teleport_to_prop(id: String, offset := Vector2.ZERO)
teleport_to_prop(id: String, offset := Vector2.ZERO)

queue_teleport_to_hotspot(id: String, offset := Vector2.ZERO)
teleport_to_hotspot(id: String, offset := Vector2.ZERO)

queue_teleport_to_marker(id: String, offset := Vector2.ZERO)
teleport_to_marker(id: String, offset := Vector2.ZERO)

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -892,7 +892,7 @@ func _teleport_to_node(node: Node2D, offset: Vector2) -> void:
 		await get_tree().process_frame
 		return
 	
-	position = node.global_position + offset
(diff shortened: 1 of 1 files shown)


mapedorr (approved):
Well coded and documented. Thank you so much!

PR #207, opened 2024-04-19
added support for removing markers using popochiu dock

anthonyirwin82:
This adds the code to correctly remove markers from the room tab using the popochiu dock


mapedorr (approved):
Well done. I didn't test Marker's deletion from the Room tab yesterday 🙃 .

PR #218, opened 2024-04-22
refs 217: added option for change spoken text or dialog option to gibberish

mefjak:
added gibberish generator

https://github.com/carenalgas/popochiu/assets/22888465/04cb617d-f63e-478d-a92c-46daf4f8f15d


addons/popochiu/editor/config/config.gd
```diff
@@ -114,6 +118,11 @@ static func is_use_translations() -> bool:
 static func get_dialog_style() -> int:
 	return _get_project_setting(DIALOG_STYLE, DialogStyle.ABOVE_CHARACTER)
 
+static func should_talk_gibberish() -> bool:
```
stickgrinder, line 121: I like the naming of these functions, but they are inconsistent with the siblings in the Config class.

We can introduce this exception but in this case, I'd consider propagating this semantics to the rest of the plugin/engine over time, so we don't have mix-and-matches, or all contributions writing functions at their taste.

@mapedorr what do you think?

addons/popochiu/engine/interfaces/i_dialog.gd
```diff
@@ -87,7 +87,25 @@ func finish_dialog() -> void:
...
+			bbcode = true
+			
+		if (!bbcode):
+			if (chr != ' '):
+					output_text += letters[randi_range(0,letters.size()-1)]
+			else:
+				output_text += ' ' 
+		else:
+			output_text += chr
+		if(chr == ']'):
```
stickgrinder, line 106: I would move this up, at line 98.

addons/popochiu/engine/objects/graphic_interface/components/dialog_text/dialog_text.gd
```diff
@@ -70,7 +70,14 @@ func _input(event: InputEvent) -> void:
...
 func play_text(props: Dictionary) -> void:
-	var msg: String = E.get_text(props.text)
+	var display_text: String = ""
+	
+	if PopochiuConfig.should_talk_gibberish():
+		display_text = D.create_gibberish(props.text)
+	else:
+		display_text = props.text
+	
+	var msg: String = E.get_text(display_text) #E.get_text(props.text)
```
stickgrinder, line 80: Not sure this makes sense.

Isn't it better to gibberish the result of `E.get_text(props.text)` instead?

stickgrinder (commented):
I've left some comments. One thing that I would ask is adding docs for this feature.

A guide on contributing docs is still missing, but you can add a page in the `how-to-create-a-game/advanced-techniques` section, by cloning one of the available markdown files and adjusting its weight (position in the side navbar) in the front matter, and adjusting the text.

Later, when the doc on the settings is complete, we'll move that section in there.

Thanks

mefjak:
cool today I'll try to make some changes :)

addons/popochiu/editor/config/config.gd
mefjak: about that function - I did follow what was in beta1 :
```
func should_remove_source_files() -> bool:
	return _get_editor_setting(_REMOVE_SOURCE_FILES_KEY, true)
```

But I can change that :)

stickgrinder: Hi @mefjak ! No need to rename it, I am dumb! Ignore my comment, and let's keep this. I also like it, as I wrote :D

Sorry

mefjak: No problem mate :)

addons/popochiu/engine/interfaces/i_dialog.gd
mefjak: the reason why is like that is that if i moved that checking car if it's ']' and change become=false then it won't add it to the text :) I can later change to something else :)

```
for chr in input_string:
		if(chr == '['):
			bbcode = true
		elif(chr == ']'):
			output_text += chr
			bbcode = false
			continue
		
		if (!bbcode):
			if (chr != ' '):
					output_text += letters[randi_range(0,letters.size()-1)]
			else:
				output_text += ' ' 
		else:
			output_text += chr
```

I need to think how to better approach this - it was just a suggestion of a change to get better with coding add-on and with codebase :)

addons/popochiu/editor/config/config.gd
mapedorr: Yes. I like (a lot) the use of the `should_` prefix for some boolean types. If sometimes I don't use it, it's because I'm afraid people will laugh at me when they read my code.

addons/popochiu/engine/objects/graphic_interface/components/dialog_text/dialog_text.gd
mapedorr: @stickgrinder is right. The result of `E.get_text` is the translation of the **String**. In the future, when we include support for gettext (PO) the thing will probably take the whole line into account to find the proper translation.

addons/popochiu/engine/interfaces/i_dialog.gd
mapedorr: I will suggest using a custom effect for **BBCode**. I did something similar [in this repo](https://github.com/Perro-Viejo/gossip_buey-prototype). We can have something like `[gibber]my text[/gibber]`. So the only thing you'll need to do is to wrap the input **String** in those tags.

https://github.com/Perro-Viejo/gossip_buey-prototype/blob/main/popochiu/graphic_interface/dialog_text/highlight_text_effect.gd

The code is not polished, but basically in this script I changed the characters by a random glyph to reveal only those that depend on the `Globals.dialog_listened_indexes` which are the indexes in the array of characters of said **String**.

mapedorr:
Hey @mefjak . Sorry for dissappearing for so long. Do you think it is possible that you put this branch up do date with the last changes in develop so we can merge it before the release of Popochiu 2? If you don't have the time to do so, I can try to help.

mefjak:
no problem :) i try to update it :)

mefjak:
@mapedorr  sorry for not keeping up-to-date. I did try to update it now. But I am still having difficult time to properly test it :)

mapedorr:
No worries. If it is fine for you, we can help with the testing.

mefjak:
If it's ok please do - but if you got time can you describe your workflow? I think I didnset up everything but always got some problems with git or testing 😅

mapedorr:
Hey @mefjak ! I made a small update on your branch to fix a bug, but appart from that, looks like it is ready to be merged. I'll make a couple of additional tests tomorrow just to double check everything is fine.

mefjak:
Cool, thank you very much

PR #223, opened 2024-04-25
Added information about default inventory icon sizes

anthonyirwin82:
Added the following as a note to the inventory documentation:

The default Popochiu GUI Templates have different inventory icon sizes. Images will be scaled to fit into the inventory slot. As a general rule it's better to scale a larger image down to a smaller one to avoid having unwanted artifacts in the image.

**2-Click Context-Sensitive:** has an inventory icon size of 16x16 pixels.
**Sierra:** has an inventory icon size of 24x24 pixels.
**9 Verb (LucasArts):** has an inventory icon size of 40x24 pixels.


mapedorr:
Thank you so much for the note!

Issue #232, opened 2024-05-04
Add support for viewport-drawn gizmos for PopochiuClickable properties


stickgrinder:
### Benefit description

Currently, the gizmos connected to PopochiuClickable objects are implemented as hidden nodes in the object themselves.

For example, the WalkTo point is chosen by a rectangle element that is selected by a button in the toolbar.

TODO: finish the description o this issue, providing animated gifs.

### Solution description

TODO: describe the solution and provide links to the source material

### Exclusions

Markers are not clickable, but they have to be positioned on the scene, so they are candidates for gizmos too.
This issue only takes into consideration PopochiuClickable objects, with gizmos for points and baselines.

### Implications

TODO

### Done when

* [x] The gizmo to position the "Walk To Point" is drawn as a rectangle on the viewport
* [x] The gizmo to position the "Baseline" is drawn as a line with a square handle on the viewport
* [x] The relevant properties for the visual representation of the gizmos are configurable as "Editor Settings":
  * [x] Size of the handles
  * [x] Colors of the handles
  * [x] Visibility of the clickable node name labels
  * [x] Outline visibility
  * [x] Connectors visibility
* [x] ~~A button is available in the Clickables toolbar to hide or show the gizmos (all of them)~~ (split in #134)
* [x] The gizmos actually update the properties they are linked to
* [x] The old gizmos implemented as node helpers should be removed from Props, HotSpots and Characters
* [x] Update documentation (change screenshot and description in "Getting Started / Create game stub" section of the docs)
* [x] Change DialogPos from Marker2D to a position gizmo
* [x] We have an upgrade path (see #227) to:
    * [x] remove the node helpers in every Popochiu object for projects that already have them
    * [x] ~~add `POPOCHIU_OBJ_POLYGON_GIZMO` meta to every polygon in the popochiu objects that have one~~ (no more relevant, we implemented this with groups set at runtime)

drbbgh:
Character properties such as `baseline` and `walk_to_point` coordinates should be visible in the inspector for character instances in the room.
Also, `walk_to_point` coordinates should be integers (see [Vector2i](https://docs.godotengine.org/en/stable/classes/class_vector2i.html)).

stickgrinder:
This became related to #134 and has been moved for release with Popochiu 2.0 (cc @drbbgh @mapedorr )

PR #241, opened 2024-05-25
refs #232: added dialog_pos management with gizmo.

drbbgh:
@mapedorr, please check that the changes to `PopochiuUtils.get_screen_coords_for()` have no unwanted impact.


stickgrinder:
@mapedorr LGTM, almost all boilerplate code, but there are small changes to the engine code you may want to review.

mapedorr (changes requested):
After testing the branch in a project created from zero, I have found a couple of issues that might be addressed before merging:

## Issues 🐞 
- If I'm editing an object's *Polygon and press **ESC** or click on another object within the Canvas Editor, the previous object's *Polygon keeps visible.

https://github.com/carenalgas/popochiu/assets/4536477/b4e30d59-a7e0-4ef5-829f-febcd18d773d

- If I change the **InteractionPolygon** of a PopochiuRegion inside the room, changes are lost when the scene is closed and opened again.

  > UPDATE: I'm also losing changes made to InteractionPolygon in props and hotspots.

## Recomendations 😳 
- How about hiding the Walk-to Point gizmo if the **PopochiuClickable** `clickable` property is `false`?
- Change the Walk-to Point icon to green. The current icon uses Cyan, which causes confusion with the baseline gizmo color. Likewise, the Baseline icon could be Cyan.
- Create an icon for the Dialog Position button. It should be Magenta to match the color of the gizmo.
- Add a configuration option (Project Settings) to define whether one wants the **PopochiuWalkableArea** polygons to always be visible.

About the icon colors, and knowing devs can customize the color of their gizmos, maybe those icons could be black and white, or in grayscale, so we can change their `modulate` property to match the colors defined in the Editor Settings.

## Question 🤔 
This can be addressed later, but I want to know your opinion.

- What if we use our custom toolbar instead of the `EditorPlugin.CONTAINER_CANVAS_EDITOR_MENU`? We have some options like: `CONTAINER_CANVAS_EDITOR_SIDE_LEFT`, `CONTAINER_CANVAS_EDITOR_SIDE_RIGHT`, and `CONTAINER_CANVAS_EDITOR_SIDE_BOTTOM`.

stickgrinder:
Changes discussed with @mapedorr and @drbbgh:

* [ ] To make sure walkable area perimeters are saved, we will test saving all of the outlines in the NavigationPolygon, then clearing, repopulating all outlines, and baking the polygon.
* [ ] If we find a way to sync the polygons and the object polygon vertices collection on scene save, we can do it outside `_process()` and spare a lot of computation. This needs to be investigated and tested.
* [ ] Switch from the meta identifying the polygons to a nodes group, so we keep things consistent with the rest of the plugin.

addons/popochiu/editor/canvas_editor_menu/popochiu_canvas_editor_menu.gd
```diff
@@ -3,114 +3,190 @@ extends HBoxContainer
...
+	):
+		return
+
+	# If we are in a room scene, we may have selected a room object of sort, so check
+	# for the various types and hide the ones we don't need
+	if PopochiuEditorHelper.is_room(EditorInterface.get_edited_scene_root()):
+		# If we are editing a clickable object, let's show gizmos buttons too
+		if _active_popochiu_object is PopochiuClickable:
+			btn_baseline.show()
+			btn_walk_to_point.show()
```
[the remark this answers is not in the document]
stickgrinder: This is now fixed: the buttons' status and gizmos should always be synced.

stickgrinder:
I'm going to answer your comments step by step so I don't forget something:

>   * If I'm editing an object's *Polygon and press **ESC** or click on another object within the Canvas Editor, the previous object's *Polygon keeps visible.

Fixed: now with esc or wild click, you go back to the object holding the polygon.

>  * If I change the **InteractionPolygon** of a PopochiuRegion inside the room, changes are lost when the scene is closed and opened again.
>       > UPDATE: I'm also losing changes made to InteractionPolygon in props and hotspots.

Everything has been fixed and I also optimized the code that saves those propertied in the PopochiuClickable. Now syncing doesn't happen on every frame but only on saving. The `editing_polygon` variable was then useless and I removed it.

> ## Recomendations 😳
> 
>     * How about hiding the Walk-to Point gizmo if the **PopochiuClickable** `clickable` property is `false`?

Considering this, we don't think this is good: in 2.0 you can change the `clickable` property by script, and you may want to set the WTP for a prop or hotspot which is not initially clickable, from the start.

We'll leave the visibility setting to the user, by mean of the toolbar buttons.

>     * Change the Walk-to Point icon to green. The current icon uses Cyan, which causes confusion with the baseline gizmo color. Likewise, the Baseline icon could be Cyan.

Checked that they don't conflict.

>     * Create an icon for the Dialog Position button. It should be Magenta to match the color of the gizmo.

Done, it was obvious but we forgot in the first iteration.

>     * Add a configuration option (Project Settings) to define whether one wants the **PopochiuWalkableArea** polygons to always be visible.

Done

> About the icon colors, and knowing devs can customize the color of their gizmos, maybe those icons could be black and white, or in grayscale, so we can change their `modulate` property to match the colors defined in the Editor Settings.

Done. There is also a preference for that behavior.

> ## Question 🤔
> 
> This can be addressed later, but I want to know your opinion.
> 
>     * What if we use our custom toolbar instead of the `EditorPlugin.CONTAINER_CANVAS_EDITOR_MENU`? We have some options like: `CONTAINER_CANVAS_EDITOR_SIDE_LEFT`, `CONTAINER_CANVAS_EDITOR_SIDE_RIGHT`, and `CONTAINER_CANVAS_EDITOR_SIDE_BOTTOM`.

We would leave the toolbar where it belongs: that's where the user goes and doesn't break the usual pattern. We added a label to tell Popochiu buttons apart from the rest :) It should work pretty fine.


With all of that, the PR should be over.
Let us know if you notice something.

addons/popochiu/editor/canvas_editor_menu/popochiu_canvas_editor_menu.gd
```diff
@@ -3,114 +3,289 @@ extends HBoxContainer
...
+	btn.add_theme_color_override("icon_hover_pressed_color", color.lightened(1.0))
+
+
+## Internal helper to reduce code duplication
+func _reset_toolbar_button_color(btn) -> void:
+		btn.remove_theme_color_override("icon_normal_color")
+		btn.remove_theme_color_override("icon_hover_color")
+		btn.remove_theme_color_override("icon_focused_color")
+		btn.remove_theme_color_override("icon_pressed_color")
+		btn.remove_theme_color_override("icon_hover_pressed_color")
```
mapedorr, line 245:
```suggestion
	btn.remove_theme_color_override("icon_normal_color")
	btn.remove_theme_color_override("icon_hover_color")
	btn.remove_theme_color_override("icon_focused_color")
	btn.remove_theme_color_override("icon_pressed_color")
	btn.remove_theme_color_override("icon_hover_pressed_color")
```

addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -20,20 +20,59 @@ extends Area2D
...
+	# When the game is running...
+	# Update the node's polygon when:
+	if (
+		get_node_or_null("InteractionPolygon") # there is an InteractionPolygon node
+	):
+		get_node("InteractionPolygon").polygon = interaction_polygon
+		get_node("InteractionPolygon").position = interaction_polygon_position
+
+
+func _notification(event):
```
mapedorr, line 71: Add the type of the `event` parameter, and the return type of the function:

```suggestion
func _notification(event: int) -> void:
```

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -107,11 +106,10 @@ func _ready():
 	_translate()
 
 
-func _process(delta):
-	if Engine.is_editor_hint():		
-		if editing_polygon:
-			interaction_polygon = get_node("InteractionPolygon").polygon
-			interaction_polygon_position = get_node("InteractionPolygon").position
+func _notification(event):
```
mapedorr, line 109: Add the type of the `event` parameter, and the return type of the function:

```suggestion
func _notification(event: int) -> void:
```

By the way, I like this approach.

addons/popochiu/engine/objects/walkable_area/popochiu_walkable_area.gd
```diff
@@ -27,19 +27,79 @@ var map_rid: RID
...
+		# Restore the NagivationRegion2D position
+		get_node("Perimeter").position = interaction_polygon_position
+
+	# Map the necessary resources
 	map_rid = NavigationServer2D.get_maps()[0]
 	rid = ($Perimeter as NavigationRegion2D).get_region_rid()
 	NavigationServer2D.region_set_map(rid, map_rid)
 
 
+func _notification(event):
```
mapedorr, line 91: Add the type of the `event` parameter, and the return type of the function:

```suggestion
func _notification(event: int) -> void:
```

mapedorr (changes requested):
Everything looks pretty good so far. I'll add the required step in Migration 2 so Popochiu objects from previous Popochiu versions are also added to the corresponding `popochiu_object_polygon`.

I asked for some minor code changes.

Now its time to test in the Engine.

stickgrinder:
@mapedorr I addressed your suggestions (god bless GitHub's suggestions).

I think I should add the clickable elements polygon to the group programmatically on _ready() and fix the references to the icons, before this can be closed, right?

mapedorr:
@stickgrinder Yes. After that I'm gonna make a couple of tests to check everything is fine and then I'll merge.

mapedorr (approved):

mapedorr:
After making some tests I found a couple of bugs and have some additional comments.

# Bugs

### Can't edit characters' polygon

When editing characters, pressing the Interaction Polygon button the Popochiu CanvasEditor menu doesn't select the **InteractionPolygon** node.

https://github.com/carenalgas/popochiu/assets/4536477/45fda27d-4767-4df3-8764-c7b9bc51ffa1

### Characters can't move in new project (and maybe in new rooms)

After creating a new project with a room, the PC didn't move, even though the walkable area is properly configured. The only difference I noticed between an older project (the editor in the background) and the new one (the editor in the foreground) is that the new one appears to have all the layers disabled in the `parsed_collision_mask` property of the NavigationPolygon.

cant_walk_on_new_project

### Can't play the game

This is due to a bug in the `editor_config.gd` script. You can still play the game by pressing F12 when the error is shown in the Debug panel.

For this I will make a code suggestion, but feel free of putting that in a separate private static function if you think is too dirty.

# To improve in later versions

When you are editing the **InteractionPolygon** of any object and click another one, the polygon of the previous one keeps visible on screen. Is not a problem from my point of view, so feel free to change this in this PR or leave it for later.

https://github.com/carenalgas/popochiu/assets/4536477/1f453287-559b-4abd-84cd-eec744f4d805

# Things to discuss

I miss being able to see the interaction polygon when selecting a Popochiu object (prop, hotspot, etc.) without needing to click the button for editing it. If this is the expected behavior, maybe devs should be able to configure whether they want the interaction polygons to be visible when selecting a Popochiu object. Or maybe in the future it will be necessary to have two buttons: one to show/hide the polygon, another to allow editing it.

PR #242, opened 2024-05-26
9 verbs interface: working GIVE command

ArturM:
On engine level Give is the same as Use, so only few changes to 9_verb_commands.gd and 9_verb_gui.gd
and no changes on the main engine.
Refactored _show_use_on() to _show_command_on() - one function for USE and GIVE and moved prepositions ('use ON', 'give TO') to one place so we can easy translate it in the future.
It calls _on_item_used() just like USE because we often wants to 'Give item to character' and 'Use item on character'  to work the same. We can check and differentiate behaviour by testing E.current_command

Like this:

```
func _on_item_used(item: PopochiuInventoryItem) -> void:
	if E.current_command == NineVerbCommands.Commands.GIVE:
		if item == I.Mushrooms:
			await C.player.say("Give mushroom")
		if item == I.Pumpkin:
			await C.player.say("Give pumpkin")
	else:
		if item == I.Mushrooms:
			await C.player.say("Use mushroom")
		if item == I.Pumpkin:
			await C.player.say("Use pumpkin")
```

### DIFF
--- a/addons/popochiu/engine/objects/graphic_interface/templates/9_verb/9_verb_commands.gd
+++ b/addons/popochiu/engine/objects/graphic_interface/templates/9_verb/9_verb_commands.gd
@@ -110,10 +110,20 @@ func pull() -> void:
 ## Called when [code]E.current_command == Commands.GIVE[/code] and [code]E.command_fallback()[/code]
 ## is triggered.
 func give() -> void:
+	await _give_or_use(give_item_to)
+
+
+## Called when [code]E.current_command == Commands.USE[/code] and [code]E.command_fallback()[/code]
+## is triggered.
+func use() -> void:
+	await _give_or_use(use_item_on)
+
+
+func _give_or_use(callback: Callable) -> void:
 	if I.active and E.clicked:
-		give_item_to(I.active, E.clicked)
+		callback.call(I.active, E.clicked)
 	elif I.active and I.clicked and I.active != I.clicked:
-		give_item_to(I.active, I.clicked)
+		callback.call(I.active, I.clicked)
 	elif I.clicked:
 		match I.clicked.last_click_button:
 			MOUSE_BUTTON_LEFT:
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/graphic_interface/templates/9_verb/9_verb_commands.gd
```diff
@@ -110,7 +110,26 @@ func pull() -> void:
...
+				# 		click inputs on clickables and inventory items. ----------------------------
+				E.current_command = (
+					I.clicked.suggested_command if I.clicked.get("suggested_command")
+					else Commands.LOOK_AT
+				)
+				
+				I.clicked.handle_command(MOUSE_BUTTON_LEFT)
+				# ----------------------------------------------------------------------------------
+	else:	
+		await C.player.say("What?")
```
mapedorr, line 132: Since this is basically doing the same as `func use()`, what do you think about unifying the logic into a single method? The only thing that would change is the method called in the first two `if` statements. It could be something like:

```
func give() -> void:
	await _give_or_use(give_item_to)


func use() -> void:
	await _give_or_use(use_item_on)


func _give_or_use(callback: Callable) -> void:
	if I.active and E.clicked:
		callback.call(I.active, E.clicked)
	elif I.active and I.clicked and I.active != I.clicked:
		callback.call(I.active, I.clicked)
	# And so on...
```


You don't have to make the change if you don't want to, its just a suggestion.

addons/popochiu/engine/objects/graphic_interface/templates/9_verb/9_verb_gui.gd
```diff
@@ -244,8 +228,14 @@ func _on_quit_pressed() -> void:
 	quit_popup.open()
 
 
-func _show_use_on(item: PopochiuInventoryItem) -> void:
-	G.show_hover_text("%s %s in" % [E.get_current_command_name(), item.description])
-
+func _show_command_on(item_1_name: String, item_2_name: String = "") -> void:
+	var preposition = "on"
+	if E.current_command == NineVerbCommands.Commands.GIVE:
+		preposition = "to"
```
mapedorr, line 234: I like this.

mapedorr (changes requested):
I added a comment suggesting a change to reduce duplicate code. You don't have to make the change if you don't want to or don't have the time. Let me know what you think so I can merge the PR.

Thanks for bringing this up.

mapedorr (approved):
Thanks for making the change. I'm gonna merge the PR.

Issue #249, opened 2024-07-29
Setting a hotspot's property Clickable to false does not take an effect


fhgaha:
### Bug description

Setting a hotspot's property Clickable to false does not take an effect

###  Steps to reproduce

1. Create a hotspot
2. In the hotspot method `_on_click()` set property `clickable = false` 
3. Watch the hotspot still being clickable in runtime even after being clicked the first time

### Expected vs observed behavior

Should become not clickable

### Environment information (please complete):

 - Windows 10
 - 4.2.2: 
 - develop branch commit:

mapedorr:
Thanks for reporting the bug. We'll give it a look.

mapedorr:
This was actually affecting all PopochiuClickable objects. It is now fixed by #254 ,

PR #254, opened 2024-08-02
Fix changing PopochiuClickable.clickable in runtime

mapedorr:
Now it has a setter method that changes the `Area2D.input_pickable` value based on the `clickable` property.

- **fix** Set right path to **PopochiuGroup** icon.


mapedorr:
Closes #249 .

Issue #244, opened 2024-07-06
Update way to define camera limits for Room


mapedorr:
### Benefit description

The way camera boundaries are defined in a room can be confusing as it requires developers to calculate the background offset relative to the game's viewport size. Instead, it should be enough to define the room size (which in most cases should match the background dimensions) so that Popochiu can take care of the calculations and set up the camera.

### Solution description

Instead of having the current:
```gdscript
@export var limit_left := INF
@export var limit_right := INF
@export var limit_top := INF
@export var limit_bottom := INF
```

Use:
```gdscript
@export var width := E.width
@export var height := E.height
```

### Exclusions

None.

### Implications

Changing this will require adding a step to Migration 2 (or Migration 3 if we decide to move this change to Popochiu 2.1) so rooms from previous versions are updated to use this approach.

mapedorr:
Closed by #257

PR #257, opened 2024-08-12
Update how camera limits are defined for rooms

mapedorr:
Now instead of using the `camera_limits` properties (which are now **deprecated**), two new properties were added to **PopochiuRoom**s: `width` and `height`. Popochiu will calculate the camera limits based on the values of those properties instead of asking devs to calculate them by hand.

The corresponding migration step was added to Migration 2 in order to ensure old projects will be updated.


mapedorr:
Closes #244

Issue #267, opened 2024-08-23
Documentation: New branding for the site and docs


stickgrinder:
### Benefit description

Papernoise worked on the new logo and color schema. We have to update the docs site (which is also the project site) to reflect these changes.

### Solution description

In scope:

* [x] apply new color scheme to the CSS
* [x] Use the logo in the sidebar instead of just the writing
* [x] Put the mascotte in the homepage

### Exclusions

Creating a documentation page with all the assets is tracked in #268

PR #269, opened 2024-08-23
refs  #267 - Project site rebranding

stickgrinder:
Everything in scope for this card has been done.

We may do more improvements in the future, but so far it works quite well on both light and dark schemas .

I also seized the opportunity to include some improvements, adding support for FontAwesome and Mermaid chars in markdown.


mapedorr (approved):
Great work! Looking forward to see this published soon.

PR #296, opened 2024-09-28
Fix vframe calculation for Prop's current sprite frame

edmundito:
This fixes a bug where the current frame was not included in the max calculation when setting a sprite's frames with vertical frames (vframes).

The change now supports both horizontal and vertical frames. When the value is set < 0, it rolls over to the max frame, and when it's increased past the max frame, it rolls over to 0.

It also adds the `total_frames` getter to `PopochiuProp`


stickgrinder:
Great addition! Thanks :heart: 

Code LGTM but I prefer @mapedorr to have a look too before bringing it in!

PR #297, opened 2024-09-28
Fix erraneous use of E instead of I when triggering inventory right click

edmundito:
Fix an error when right-clicking on an inventory item: Instead of using the `I` singleton, it was using `E`.


stickgrinder:
Thanks @edmundito! Merging

PR #298, opened 2024-09-28
Feature: Flip dialog portrait avatar based on character direction

edmundito:
This change flips the avatar image of the character based on the character's direction.

## Work in Progress

This probably needs a bit more work before it's ready for a final review. I will add a checklist here:
- [ ]  Avoid the private character var `_looking`
- [ ] Let the developer determine whether to flip the avatar
- [ ] Let the developer set the avatar portrait direction

### DIFF
--- a/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
+++ b/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
@@ -5,6 +5,11 @@ extends PopochiuDialogText
 @onready var right_avatar_container: PanelContainer = %RightAvatarContainer
 @onready var right_avatar: TextureRect = %RightAvatar
 
+const _LOOKING_LEFT_DIRS := [
+		PopochiuCharacter.Looking.LEFT,
+		PopochiuCharacter.Looking.UP_LEFT,
+		PopochiuCharacter.Looking.DOWN_LEFT,
+		]
 
 #region Godot ######################################################################################
 func _ready() -> void:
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
```diff
@@ -30,12 +35,16 @@ func _update_avatar(chr: PopochiuCharacter, _msg := '') -> void:
 		E.scale if E.settings.scale_gui else Vector2.ONE
 	)
 	
+	var flip_h := _LOOKING_LEFT_DIRS.has(chr._looking_dir)
```
edmundito, line 38: It looks like `_looking_dir` is being misused in other areas as well and should either be a getter/setter or a public variable.

stickgrinder: Yep, I guess that's very legacy code that nobody touched in a while. Nice spot!

stickgrinder:
I like this one a lot! Have you got an idea on the API functions to expose for avatar direction management, already?
I would discuss them here so we can decide on their signatures before implementing them :)

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
```diff
@@ -5,6 +5,11 @@ extends PopochiuDialogText
 @onready var right_avatar_container: PanelContainer = %RightAvatarContainer
 @onready var right_avatar: TextureRect = %RightAvatar
 
+const _LOOKING_LEFT_DIRS := [
```
mapedorr, line 8: Following [Godot's style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#constants-and-enums), constants should be defined without underscore on the first character.

```suggestion
const LOOKING_LEFT_DIRS := [
```

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
mapedorr: You're right @edmundito . We can make that property public (by removing the underscore at the beggining) or use getter/setter methods. I prefer the first option.

Changing that will also affect the `i_room.gd` and the `popochiu_room_data.gd` scripts. And this line of code, of course.

mapedorr:
Interesting feature.

I’m afraid relying solely on whether it’s facing left might not be enough. Maybe (and this can be part of that API @stickgrinder is wondering about) characters should have a function to know if their Sprite is flipped or not, and based on that, the avatar will flip too (if it’s set up to flip with the character).

stickgrinder:
Hi @edmundito!

Did you, by any chance, managed to do more work on this PR?
If you don't plan to do more work on that, me or @mapedorr can take this over. 

Let us know, thanks.

edmundito:
I have not and may not have the time. Feel free to take over!

stickgrinder:
Thanks @edmundito!

We'll add you as reviewer, should you have any comments or suggestion (since you actually needed this feature)!

stickgrinder:
@edmundito @mapedorr I worked on another branch to address this and merged it (see #449).

@edmundito I credited you in the commit message because I started from your code, so THANKS.

I'll close this PR as soon as we release the next version, as usual. We'll test the solution as soon as @mapedorr will finish working on the new hi-res GUIs.

edmundito:
Glad to see this got built! Glad that I could grease up the wheels.

Issue #138, opened 2023-12-20
Aseprite importer: Add importer to Inventory Items


stickgrinder:
### Benefit description

Related to #8 

A specific importer must be available for Inventory Items too.

### Solution description

All inventory items can be depicted in a single Aseprite file. Every item sprite (or animation) will be tagged with its name in the same format described in #8.

Importing the file will create a separate inventory item for each tag.

### Exclusions

At the moment, each item will have a single animation. Multiple animations per item can be added later.

### Implications

It's important to note that inventory items have no sprites, but TextureRects. This is due to specific features of TextureRects that @mapedorr had to bake into the inventory GUI. If migrating to sprites can be considered, we may do it.
If not we have to add a variation on the importer to support populating TextureRects too. Reference code is already available.

PR #299, opened 2024-09-28
Feature: Add AnimatedSprite2D support to characters

edmundito:
This supports `AnimatedSprite2D` for character animation as used in my game [With Our Eyes Wide Open](https://edmundito.itch.io/with-our-eyes-wide-open). It supports anti-glide animation and ignores the Animation Player node.

image

What the user needs to do:
1. Replace the type of the `Sprite2D` node with `AnimatedSprite2D`
2. Add animations
3. Optional: Remove the AnimationPlayer node (currently, it ignores it if it detects the AnimatedSprite2D type)

## TODO
- [ ] Add edge case checks
- [ ] Code cleanup
- [ ] Discuss additional features
- [ ] Add documentation

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -103,7 +103,7 @@ var default_scale := Vector2.ONE
 
 var _looking_dir: int = Looking.DOWN
 
-@onready var _animated_sprite: AnimatedSprite2D = ($Sprite2D as AnimatedSprite2D) if $Sprite2D.get_class() == "AnimatedSprite2D" else null
+@onready var _animated_sprite: AnimatedSprite2D = $Sprite2D if $Sprite2D is AnimatedSprite2D else null
 
 
 
@@ -120,10 +120,8 @@ func _ready():
 	else:
 		set_process(follow_player)
 	
-	for child in get_children():
-		if not child is Sprite2D:
-			continue
-		child.frame_changed.connect(_update_position)
+	for sprite_child in get_children().filter(func(child): return child is Sprite2D or child is AnimatedSprite2D):
+		sprite_child.frame_changed.connect(_update_position)
 	
 	move_ended.connect(_on_move_ended)
 
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -103,6 +103,8 @@ var default_scale := Vector2.ONE
 
 var _looking_dir: int = Looking.DOWN
 
+@onready var _animated_sprite: AnimatedSprite2D = ($Sprite2D as AnimatedSprite2D) if $Sprite2D.get_class() == "AnimatedSprite2D" else null
+
 
 
 #region Godot ######################################################################################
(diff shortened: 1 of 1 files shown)


stickgrinder:
Hi @edmundito!

Adding my 2 cents (more of a monday morning braindump), with some considerations and context.

Your PR may be good as it is, in that it provides an automatic management of that node, for those who introduce it by hand, but if it stays like this, I see it more as a.. "ghost feature"? 
What I mean is that it has to be clear that if one switch to AnimatedSprite2D, it gets basic support but misses the connection to other Popochiu features:

From the top of my head I see impacts on:

1. **Aseprite Importers**: they expect the character to have a specific structure, so if they have to support many different ones, we need to create specialized classes for that (I will to that already for #138, supporting TextureRect, but for the character we may need a bit more complex strategy.

2. **User experience / options**: If we give more options on how a character is structured, we may want to provide choice to a beginner users on which type of character they wants to use in their games, either with project settings or a selection interface, or whatever. Speaking of which, can I ask why you switched to AnimatedSprite2D in the first place? (not a passive-aggressive question, I'm really interested in understanding the use-case and expected benefit)

3. **Upcoming features**: in the plans we have guides and probably a dedicated editor UI to support functions calling in animations. Think of playing a "bang" sound during a gunshot, or triggering events when an animation reaches a certain point... this comes pretty easy with AnimationPlayer, that's the reason we introduced it (and it's complexity) in the first place. I'm no expert with all Godot nodes but at a first glance having such features on top of an AnimatedSprite2D would be a lot of work.

As I said, that doesn't imply any judgment, it's just adding to the context. I'd really love to understand your use-case if you don't mind.

Thanks!

mapedorr:
Hi @edmundito and @stickgrinder . I'll add my 2 cents to the discussion.

I think supporting **AnimatedSprite2D** nodes in characters (and maybe props) is a good addition. A lot of **animations** are more easier to implement when one uses this node. And I see this specially useful for HD graphics, people using tools such TexturePacker, and beginners who don't like pixel-art.

I agree with @stickgrinder about the additional work it might require when thinking about UX, but if we want Popochiu to be the go-to tool for making 2D point 'n' click games in Godot (and maybe 3D in the future), we'll have to do our best to allow devs to work in whatever way is most comfortable for them. And if using the **AnimatedSprite2D** node complicates things too much, we can present it as a form of animation where developers will need to do additional things to take full advantage of all the features the plugin could potentially offer. Documentation will be our ally on this.

## About the implementation

I don't know if changing the type of the **Sprite2D** node is needed (in fact, I'd rather we didn't do it). We can still provide a structure for characters (and props) where one can have a **Sprite2D** and an **AnimationPlayer** node along with a **AnimatedSprite2D** one. We can add properties to those classes so devs can select in the Inspector the way animations will be played (using the AnimationPlayer+Sprite2D, or the AnimatedSprite2D).

Another option is to keep the structure as it is, and play animations by default in the AnimatedSprite2D node if the scene has one (this way devs will have to manually add the node to the tree in case they want to use it).

stickgrinder:
Hi @edmundito.

Yesterday me and @mapedorr had the time to discuss over this a bit and our position is that supporting many different cases is the best option. I'll spare you all the considerations on why.

If we guessed correctly, the main reason you decided to ditch the AnimationPlayer is that, not using aseprite and going with an exported spritesheet, it's more painful to configure than the AnimatedSprite2D. Do you confirm this, or were there different or additional reasons for your choice?

If that's the case, we think the right approach would be for the plugin to provide a "spritesheet importer" tool, that sets up the AnimationPlayer with a few clicks.
This would ease the work of those who don't use Aseprite, maintaining the engine sleek and easier to maintain.

Anyway, we are thinking about closing this PR and accepting your contributions on the docs (if you feel like writing a guide on how to customize the Character as you did in _WOEWO_) or in defining a new feature of the plugin that would support your use-case.

Let us know, thanks!

edmundito:
> If we guessed correctly, the main reason you decided to ditch the AnimationPlayer is that, not using aseprite and going with an exported spritesheet, it's more painful to configure than the AnimatedSprite2D. Do you confirm this, or were there different or additional reasons for your choice?

* We were not using asesprite for import
* Frames were exported with TexturePacker
* We were using a plugin to import the texture packer sheet (https://github.com/CodeAndWeb/texturepacker-godot-plugin)
* Our artist is used to delivering at a fixed set of framerate, so using the AnimatedSprite2D to drag and drop the imported frames from the TexturePacker plugin import was more straightforward than learning how to do this automatically with AnimationPlayer

stickgrinder:
>     * We were not using asesprite for import
> 
>     * Frames were exported with TexturePacker
> 
>     * We were using a plugin to import the texture packer sheet (https://github.com/CodeAndWeb/texturepacker-godot-plugin)
> 
>     * Our artist is used to delivering at a fixed set of framerate, so using the AnimatedSprite2D to drag and drop the imported frames from the TexturePacker plugin import was more straightforward than learning how to do this automatically with AnimationPlayer

OK, thanks!

I'll take this into an issue and make it available to you so we can come up with a solid solution that eases the process of configuring the animation player from sources different from aseprite.

I'll keep you in the loop.

PR #300, opened 2024-09-28
Feature: Show system text in queue when text does not contain colon

edmundito:
This adds the ability to show system text when the text does not contain a colon. For example:
```gdscript
await E.queue([
  "Ego: I said something.",
  "Ego raises their hands.", # This be shown as system text
  "Ego: I AM HAPPY!"
]
```

### DIFF
--- a/addons/popochiu/engine/helpers/popochiu_characters_helper.gd
+++ b/addons/popochiu/engine/helpers/popochiu_characters_helper.gd
@@ -41,7 +41,7 @@ static func execute_string(text: String) -> void:
 	elif ":" in text:
 		await _trigger_dialog_line(text)
 	else:
-		await E.get_tree().process_frame
+		await G.show_system_text(text)
 	
 	E.auto_continue_after = -1.0
 


stickgrinder:
Interesting take.

@mapedorr my memory is terrible... are there more options to show an anonymous text? If so, I would just add an option to chose the recipient in the project settings.

mapedorr:
I like what @edmundito proposes here.

The only one is `G.show_system_text()` @stickgrinder . We can discuss about that option in the **Project Settings** on a separate issue, maybe? So far, what I understand, is that you propose to allow devs to choose which character (if any) will say those lines queued without specifying the character name, am I right?.

stickgrinder:
> I like what @edmundito proposes here.
> 
> The only one is `G.show_system_text()` @stickgrinder . We can discuss that option in **Project Settings** in a separate issue, maybe? So far, what I understand is that you're proposing to allow devs to choose which character (if any) will say those queued lines without specifying the character's name, am I right?

No, I was only wondering if there are other functions to display text other than `C.name.say()` and `E.show_system_text()`. In that case an option to route the anonymous text was a plus.

Since that's not the case, I would merge this right away! :)

PR #310, opened 2024-10-17
Fixed bug with using Inventory Items so the item is no longer active after you use it

anthonyirwin82:
In point and click adventure games when you select an inventory item and click to use the inventory item with clickable objects in the game or with other inventory items, the inventory item should no longer be active as the user has already used the item and the game has either done something or displayed a message saying they can't do that.

The sierra and 9 verb templates seem to take care of it as part of the templates.

The simple two click template games was not working as expected, so I fixed it to work as described above.

### DIFF
--- a/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
+++ b/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
@@ -249,6 +249,8 @@ func on_middle_click() -> void:
 ## Called when an [param item] is used on this object.
 func on_item_used(item: PopochiuInventoryItem) -> void:
 	await _on_item_used(item)
+	# after item has been used return to normal state
+	I.active = null
 
 
 ## Triggers the proper GUI command for the clicked mouse button identified with [param button_idx],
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -348,6 +350,8 @@ func _on_input_event(_viewport: Node, event: InputEvent, _shape_idx: int):
 		MOUSE_BUTTON_LEFT:
 			if I.active:
 				await on_item_used(I.active)
+				# after item has been used return to normal state
+				I.active = null
```
mapedorr, line 354: Is this needed? You're already changing the current active inventory item inside the `on_item_used` function.

addons/popochiu/engine/objects/inventory_item/popochiu_inventory_item.gd
```diff
@@ -352,6 +354,8 @@ func _on_gui_input(event: InputEvent) -> void:
 		MOUSE_BUTTON_LEFT:
 			if I.active:
 				await on_item_used(I.active)
+				# after item has been used return to normal state
+				I.active = null
```
mapedorr, line 358: Is this needed? You're already changing the current active inventory item inside the `on_item_used` function.

mapedorr (commented):
Just asked something before approving.

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
anthonyirwin82: No it is not needed I just checked it and on_item_used() is called just before the I.active = null so it is not required. I removed the line and tested and it works as expected.

PR #311, opened 2024-10-19
closes #289: make the transition layer optional in the first room

### LINKED ISSUE #289: Make the transition layer optional in the first room
### What's the expected benefit of the feature you are requesting?

The `_on_room_entered()` function is called on the room change, and happens **before** the transition layer starts to fade away. This is useful to mask all the setup actions, like moving a character to a certains spot, making some animation play to a certain point, or whatever can "take screen time" and create visual glitches.

The problem is, the transition layer is not applied to the first room, so there is no way to perform some preparatory actions "behind the curtain" if they happen in the first room.

The absence of the TL is by design, to _save dev time_ during tests, and considering that the first room is usually a splash.

In practice this effect is very mitigated and clearly nobody notices as of today, but me coding the sample game.  
Still this is a problem that a wider adoption may arise, and the reason is, you can't trust the `_on_room_entered()` function to behave consistently. The first case out in the wild will result in a bug report.

### Describe the solution you'd like

Simple and effective solution is to add (and document) an option in the _Project Settings_, called "Fade in at game start".  
Devs will decide the behavior of their game, not being forced into anything.

### Describe alternatives you've considered

* **Changing the comment above the `_on_room_entered()` function** to explain that it won't apply to the first room (this is more of a workaround, really)
* **Changing the moment the game engine calls the function, so it happens before the room is actually displayed.** This may be unachievable in practice.
* **Making sure the room is "covered in black" until the function is over**, then removing the curtain and letting the TL work as it is working now. This will mask the room no matter what during the execution of that function, independent on how the room is faded in (fast or slow, yes or not... whatever). This may be desireable anyway, even if we add the proposed setting... maybe? In doubt we should apply the KISS principle.

stickgrinder:
The commit in this PR is already working.

Before finishing this PR I'll also try to do a slight refactor of the Transition Layer, to use ColorRects _en lieu_ of a screen-wide PNG, to achieve easier full-covering at every resolution and get rid of an external resource.


stickgrinder:
@mapedorr this is ready for review. The last commit simplifies the TL scene to use only one ColorRect. The input is masked by the TL itself that's now a control.

We passed from 5 animations to 3, playing them backwards preserves the same behavior. I tested the TL with skips and cutscenes and it works.

Also, of couse, this PR implements the masking option for the first room.

addons/popochiu/engine/objects/transition_layer/transition_layer.gd
```diff
@@ -22,62 +22,77 @@ enum {
...
+			$AnimationPlayer.play("fade")
+			await $AnimationPlayer.animation_finished
+			$AnimationPlayer.play_backwards("fade")
 			await $AnimationPlayer.animation_finished
-			
-			$AnimationPlayer.play("fade_out")
+			_hide()
 		FADE_IN:
-			$AnimationPlayer.play("fade_in")
+			print("Fading in")
```
mapedorr, line 58: Remove this.

addons/popochiu/engine/objects/transition_layer/transition_layer.gd
```diff
@@ -22,62 +22,77 @@ enum {
...
+			$AnimationPlayer.play("fade")
 		FADE_OUT:
-			$AnimationPlayer.play("fade_out")
+			print("Fading out")
```
mapedorr, line 61: Also remove this.

mapedorr (changes requested):
Great refactor. Just asked to remove a couple of prints.

mapedorr (approved):
Terrific.

Issue #315, opened 2024-10-26
Cursor incorrectly reset after invalid examine operation with 2 button interface


balloonpopper:
### Bug description

The _on_unblocked() function in popochui_gui,gd resets the cursor back to the standard cursor when it shouldn't using the 2 button interface

###  Steps to reproduce

1. Set your project up to use the 2 button interface
2. Create a prop. Set up the interaction polygon.
3. Give it an on_click function - e.g.
```
func _on_click() -> void:
	await C.player.walk_to_clicked()
```	
4. Do **not** change an _on_right_click()
5. Play your game. When you move your mouse over the interaction polygon it will change to the hotspot cursor.
6. Right click to examine and you get the default "can't EXAMINE this" error. Leave your mouse over the interaction polygon.
7. Left click. As you're still over the interaction polygon, the cursor should be the hotspot cursor. Instead it has changed back to the default cursor.

### Expected vs observed behavior

If the cursor is still over a hotspot after a failed click option, the cursor should remain as the hotspot cursor.
I expect that the cursor would also reset if you held an inventory item.

### Environment information (please complete):

 - Windows 10
 - 4.2
 - Popochui v2.0.0

mapedorr:
Looks like something easy to fix. I'll give it a look. Thanks for reporting it!

mapedorr:
Closed by #323

PR #323, opened 2024-10-30
refs #315: Show the proper cursor after closing SystemText

mapedorr:
This occurs when closing the SystemText component while the cursor is over a PopochiuClickable object.


mapedorr:
Closes #315

PR #324, opened 2024-10-30
Fix #320 by ignoring objects that are not part of the room tree when creating its data resource

### LINKED ISSUE #320: Issue when changing to a room with a missing hotspot
### Bug description

As discussed here on Discord :
https://discordapp.com/channels/970008012553068583/970133408737005588/1300088547868414003
When changing rooms to a room with a broken hotspot (i.e. you've deleted the hotspot but not the files on the filesystem relating to the hotspot), the game breaks and "Cannot call method 'get_script' on a null value" is displayed. 
Error is on popochui_resources.gd:474, from i_room.gd:270, from popochui_room.gd:109

When you change rooms, the code in i_room.gd searches through all the hotspots in your room. If a hotspot has been deleted but the files still exist on the filesystem, the error happens. The strange thing about it is that the room you start the game from is not evaluated.

###  Steps to reproduce

In both scenarios, room 1 is the default game start room. Hotspots are broken by creating a hotspot, deleting the hotspot, but leaving the files on the filesystem.

Scenario 1 :  I break a hotspot in room 1. If I start the game from room 1 (F5 or F6), I can go between room 1 and 2 as many times as I like and it won't break. If I start the game from room 2 (F6) and change to room 1, it breaks.

Scenario 2 : I break the a hotspot in room 2 instead. If I start the game from room 2 (F6), I can go between room 2 and room 1 as many times as I like and it won't break. If I start the game from room 1 (F5 or F6), when I go to room 2 it breaks.

The engine should print a debug error about a missing hotspot but continue, instead it breaks.

 - Windows 10
 - Godot 4.4 dev 2
 - Popochui 2.0.0

### Additional information

The code below (null check on the last 2 lines) fixes the problem, but I'm not submitting it as a pull request as it appears to be more of a mask for the problem than a fix. The package maintainers may decide it's sufficient to fix the problem though.

i_room.gd : line 259
```
	for type in PopochiuResources.ROOM_CHILDREN:
		print("RS = "+str(rooms_states[room.script_name][type]))
		for script_name in rooms_states[room.script_name][type]:
			var node: Node2D = current.callv(
				"get_" + type.trim_suffix("s"),
				[(script_name as String).to_pascal_case()]
			)
			if node == null:
				continue
```

mapedorr:
This takes into account the dependencies of the room to avoid loading objects to its data resource when they are not part of the room tree.

### DIFF
--- a/addons/popochiu/engine/objects/room/popochiu_room_data.gd
+++ b/addons/popochiu/engine/objects/room/popochiu_room_data.gd
@@ -1,16 +1,16 @@
-@icon('res://addons/popochiu/icons/room.png')
+@icon("res://addons/popochiu/icons/room.png")
 class_name PopochiuRoomData
 extends Resource
 ## This class is used to store information when saving and loading the game. It also ensures that
-## the data remains throughout the game's execution.
+## the data remains throughout the game"s execution.
 ##
 ## It also has data of the [PopochiuProp]s, [PopochiuHotspot]s, [PopochiuWalkableArea]s,
 ## [PopochiuRegion]s, and [PopochiuCharacter]s in a [PopochiuRoom].
 
 ## The identifier of the object used in scripts.
-@export var script_name := ''
+@export var script_name := ""
 ## The path to the scene file to be used when adding the character to the game during runtime.
-@export_file("*.tscn") var scene := ''
+@export_file("*.tscn") var scene := ""
 ## Whether the room was already visited by the player.
 @export var visited := false
 ## Whether this is the first time the player visits the room.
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/room/popochiu_room_data.gd
```diff
@@ -1,16 +1,16 @@
-@icon('res://addons/popochiu/icons/room.png')
+@icon("res://addons/popochiu/icons/room.png")
 class_name PopochiuRoomData
 extends Resource
 ## This class is used to store information when saving and loading the game. It also ensures that
-## the data remains throughout the game's execution.
+## the data remains throughout the game"s execution.
```
stickgrinder, line 5: waaaaiiiit a minute!!! :D 

```suggestion
## the data remains throughout the game's execution.
```

addons/popochiu/engine/objects/room/popochiu_room_data.gd
```diff
@@ -162,7 +176,7 @@ func _save_object_state(node: Node2D, ignore: Array, target: Dictionary) -> void
 	var state := {}
 	PopochiuResources.store_properties(state, node, ignore)
 	
-	# Add the PopochiuProp state to the room's props
+	# Add the PopochiuProp state to the room"s props
```
stickgrinder, line 179:
```suggestion
	# Add the PopochiuProp state to the room's props
```

stickgrinder (changes requested):

Issue #331, opened 2024-11-07
Gibberish Spoken Text toggle in settings not working


el-falso:
### Bug description

The toggle in the Project Settings "Gibberish Spoken Text" isn't working.

###  Steps to reproduce

1. Turn "Gibberish Spoken Text" in the Project Settings on
2. Let a character in the Game say something
3. You see normal text when you start the game

### Expected vs observed behavior

I expected the spoken text to be gibberish, but instead it's the text you wrote in your code.

### Environment information (please complete):

 - **OS type and version**: Manjaro KDE Plasma
 - **Godot version**: 4.3
 - **Plugin version or branch**: 2.0.1

PR #332, opened 2024-11-07
fix: gibberish in spoken texts not working

el-falso:
Somehow the "Gibberish Spoken Text" option in the Project Settings isn't used in Popochiu in any way. This PR fixes this, so when you toggle this setting on, characters speak Gibberish instead.

Fixes: #331


mapedorr (approved):
Thanks for fixing this!

PR #339, opened 2024-11-11
issue_334 Better direction choices based on angle

balloonpopper:
Corrects unexpected walking angles seen in the previous commit based off my testing.
Also preferences more animation angles so it should handle 8, 4 or 2 direction animations.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -758,69 +758,54 @@ func face_direction(destination: Vector2):
 	# right. Flipping is done in other functions. We just define
 	# a preference order for animations when available.
 
-
-	if angle < 180:
-		if angle < 90:
-			if angle < 45:
-				if angle < 22.5:
-					_animation_suffixes = ['_r', '_dr', '_l', '_dl', '_d']
-					_looking_dir = Looking.RIGHT
-				else:
-					_animation_suffixes = ['_dr', '_r', '_dl', '_l', '_d']
-					_looking_dir = Looking.DOWN_RIGHT
-			else:
-				if angle < 67.5:
-					_animation_suffixes = ['_dr', '_d', '_dl', '_r', '_l']
-					_looking_dir = Looking.DOWN_RIGHT
-				else:
-					_animation_suffixes = ['_d', '_dr', '_dl', '_r', '_l']
-					_looking_dir = Looking.DOWN
-		else:
-			if angle < 135:
-				if angle < 112.5:
-					_animation_suffixes = ['_d', '_dl', '_dr', '_l', '_r']
-					_looking_dir = Looking.DOWN
-				else:
-					_animation_suffixes = ['_dl', '_d', '_dr', '_l', '_r']
-					_looking_dir = Looking.DOWN_LEFT
-			else:
-				if angle < 157.5:
-					_animation_suffixes = ['_dl', '_l', '_dr', '_r', '_d']
-					_looking_dir = Looking.DOWN_LEFT
-				else:
-					_animation_suffixes = ['_l', '_dl', '_r', '_dr', '_d']
-					_looking_dir = Looking.LEFT
+	if angle < 22.5:
+		_animation_suffixes = ['_r', '_l', '_dr', '_dl', '_d']
+		_looking_dir = Looking.RIGHT
+	elif angle < 45:
+		_animation_suffixes = ['_dr', '_dl', '_r' , '_l', '_d']
+		_looking_dir = Looking.DOWN_RIGHT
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -297,7 +297,7 @@ func queue_face_up_right() -> Callable:
 ## Makes the character to look up and right by setting [member _looking_dir] to [constant UP_RIGHT]
 ## and waits until [method idle] finishes.
 func face_up_right() -> void:
-	_looking_dir = Looking.UP_RIGHT
+	face_direction(position + Vector2(10, -10))
```
stickgrinder, line 300: This is clever but it can be more self-explanatory.

Vector2 has some constants, like `Vector2.RIGHT` and `Vector2.UP`. What about using

```suggestion
	face_direction(position + Vector2.UP + Vector2.RIGHT
```

or

```suggestion
	face_direction(position + (Vector2.UP + Vector2.RIGHT) * 10
```

It's just a small addition, but it makes it very readable.

balloonpopper: I like that solution much better!

balloonpopper:
@stickgrinder I agree that yours is more readable. I was trying to optimise it from (worst case) 16 if statements down to 4, but happy to prioritise readability / maintainability. Modified accordingly.

I made some changes to my previous commit to my fallback arrays, I believe they are correct now.
I think there's some differences in my preference choices, and also some bugs in the original code.

I've only traced out the first section of the original code, this may be the only bug, or the might be more.
Below is the code from L765 to L770 with the variables replaced by the values to make it easier to read:
https://github.com/carenalgas/popochiu/commit/80c645999e14ec7d2960dbe716b3fc70aa40e6c2#diff-099209ce75040aaa85afb7f28211f97fbbaa3499bb39c0e9dae1b98df6705da7R765
```
	elif angle >= (22.5) and angle < (67.5):
		_looking_dir = Looking.DOWN_RIGHT
		if angle >= (45) and angle < (135):
			_animation_suffixes = ['_dr', '_r', '_dl', '_l']
		else:
			_animation_suffixes = ['_dr', '_dl', '_d']
```
This check "if angle >= (45) and angle < (135):" will work, but the 135 is the wrong value. If anything it should've been 67.5 - but only the >=45 check is actually needed.

Then in relation to my angle preferences vs yours:
For an angle of 60 degrees, the original code shown above will preference R if DR is unavailable (['_dr', '_r', '_dl', '_l']). 60 is closer to 90 so it should prefer down.

addons/popochiu/engine/objects/character/popochiu_character.gd
balloonpopper: Updated - great idea!

balloonpopper:
Additionally in relation to : https://github.com/carenalgas/popochiu/issues/340
As a test I added this code after the rad_to_deg angle is calculated:
```
        # Top of file has "var _last_direction_faced:int = -1"
        ########################################

	var direction_faced = int(angle/45)
	if direction_faced == _last_direction_faced:
		return
	_last_direction_faced = direction_faced
```
It fixes the code having to constantly reevaluate the facing direction, but I believe the whole walking subsystem should likely be optimised (see issue 340). 
If you don't agree with 340, but think the above code is helpful, I can add it to the PR.

stickgrinder:
> This check "if angle >= (45) and angle < (135):" will work, but the 135 is the wrong value. If anything it should've been 67.5 - but only the >=45 check is actually needed.

You are right about that being a problem; the original code was very hard to read and it's way better now anyway, thanks!

> Then in relation to my angle preferences vs yours: For an angle of 60 degrees, the original code shown above will preference R if DR is unavailable (['_dr', '_r', '_dl', '_l']). 60 is closer to 90 so it should prefer down.

That's ok to me! Being a matter of preferences, the best option is trying it in the wild and see if someone raise some complaints :) If not, your version _just works_ :magic_wand: 

About #340, I totally agree. That code is very legacy and my personal take is that it grew over time and has never been re-evaluated because to a modern machine that's trivial stuff anyway. **_BUT_** that's not a good reason to keep inefficiency in code, so: great catch.

What I would do, if you agree is: review and merge this PR as it is, if you consider it ready, and ask you to open a dedicated one for #340 with all the optimization you mentioned.

What do you think?

balloonpopper:
Happy for you to merge the code as is if you're happy with it. It has the history if you want this discussion to be able to refer back to how it developed. 
If you want to keep the commits down though let me know and i will squash it to a single commit.

stickgrinder:
No need to squash manually, gh can do this during the merge.

This discussion makes enough sense tracing back the most important parts :) Thanks!

Issue #340, opened 2024-11-12
Optimising walking processing


balloonpopper:
### What's the expected benefit of the feature you are requesting?

While stepping through the code to find a bug, I've found that walking seems to be very inefficient code-wise.
Every frame
* The angle the character is facing is rechecked
* The play_animation function loops through every animation suffix checking if it has an animation for the direction the character is facing
* If the character isn't on a scaling region its walk_speed is reset

All of this seems very inefficient (plus it makes debugging hard as you step through lots of (potentially) unnecessary code)

### Describe the solution you'd like

Walk speed/direction/animation being used etc are all cached. If something changes they're reevaluated. The navigation should be something like
If the character isn't on a scaling region, and they're travelling the same direction they were last time, reuse the same walk speed and animation and move the character.

### Describe alternatives you've considered

None

### Additional context

None

PR #341, opened 2024-11-13
issue_340 Add caching to animation player and navigation to avoid unnecessary calls

balloonpopper:
See https://github.com/carenalgas/popochiu/issues/340

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -15,11 +15,6 @@ enum FlipsWhen {
 }
 ## Determines the direction the character is facing
 enum Looking {
-	## The character is facing up [code](0, -y)[/code].
-	UP,
-	## The character is facing up-right [code](x, -y)[/code].
-	UP_RIGHT,
-	## The character is facing right [code](x, 0)[/code].
 	RIGHT,
 	## The character is facing down-right [code](x, y)[/code].
 	DOWN_RIGHT,
@@ -30,7 +25,12 @@ enum Looking {
 	## The character is facing left [code](-x, 0)[/code].
 	LEFT,
 	## The character is facing up-left [code](-x, -y)[/code].
-	UP_LEFT
+	UP_LEFT,
+	## The character is facing up [code](0, -y)[/code].
+	UP,
+	## The character is facing up-right [code](x, -y)[/code].
+	UP_RIGHT
+	## The character is facing right [code](x, 0)[/code].
 }
 
 ## Emitted when a [param character] starts moving from [param start] to [param end]. [PopochiuRoom]
@@ -119,6 +119,24 @@ var _last_animation_direction: int = -1
 
 @onready var animation_player: AnimationPlayer = $AnimationPlayer
 
+# Array of the animation suffixes to prefer based on the angle the character is facing
+var _valid_animation_suffixes = [
+['_r', '_l', '_dr', '_dl', '_d'], #    0 - 22.5 degrees
+['_dr', '_dl', '_r' , '_l', '_d'], #  22.5 - 45 degrees
+['_dr', '_dl', '_d' , '_r', '_l'], #  45 - 67.5 degrees
+['_d', '_dr', '_dl', '_r', '_l'], #   67.5 - 90 degrees
+['_d', '_dl', '_dr', '_l', '_r'], #  90 - 112.5 degrees
+['_dl', '_dr', '_d', '_l', '_r'], # 112.5 - 135 degrees
+['_dl', '_dr', '_l', '_r', '_d'], # 135 - 157.5 degrees
+['_l', '_r', '_dl', '_dr', '_d'], # 157.5 - 180 degrees
+['_l', '_r', '_ul', '_ur', '_u'], # 180 - 202.5 degrees
+['_ul', '_ur', '_l', '_r', '_u'], # 202.5 - 225 degrees
+['_ul', '_ur', '_u', '_l', '_r'], # 225 - 247.5 degrees
+['_u', '_ul', '_ur', '_l', '_r'], # 247.5 - 270 degrees
+['_u', '_ur', '_ul', '_r', '_l'], # 270 - 292.5 degrees
+['_ur', '_ul', '_u', '_r', '_l'], # 292.5 - 315 degrees
+['_ur', '_ul', '_r', '_l', '_u'], # 315 - 337.5 degrees
+['_r', '_l', '_ur', '_ul', '_u']] # 337.5 - 360 degrees
 
 #region Godot ######################################################################################
 func _ready():
@@ -231,7 +249,6 @@ func queue_walk(target_pos: Vector2) -> Callable:
 func walk(target_pos: Vector2) -> void:
 	is_moving = true
 	_last_reached_clickable = null
-	_looking_dir = Looking.LEFT if target_pos.x < position.x else Looking.RIGHT
 
 	# Make the char face in the correct direction
 	face_direction(target_pos)
@@ -753,16 +770,6 @@ func resume_animation():
 ## Makes the character look in the direction of [param destination]. The result is one of the values
 ## defined by [enum Looking].
 func face_direction(destination: Vector2):
-	# Get the vector from the origin to the destination.
-
-	var angle = wrapf(rad_to_deg((destination - position).angle()), 0, 360)
-	var direction_faced = int(angle/45)
-	if direction_faced == _last_looking_dir:
-		return
-	_last_looking_dir = direction_faced
-	# Tolerance in degrees, to avoid U D L R are only
-	# achieved on precise angles such as 0 90 180 deg.
-
 	# Determine the direction the character is facing.
 	# Remember: Y coordinates have opposite sign in Godot.
 	# this means that negative angles are up movements.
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -108,6 +108,14 @@ var _looking_dir: int = Looking.DOWN
 var _animation_suffixes: Array = ['_d', '_l', '_r', '']
 # Holds the last PopochiuClickable that the character reached.
 var _last_reached_clickable: PopochiuClickable = null
+# Caches the last direction the character faced to avoid unnecessary recalculation
+var _last_direction_faced: int = -1
```
stickgrinder, line 112: I thinks this can be called `_last_looking_dir' for consistency?

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -108,6 +108,14 @@ var _looking_dir: int = Looking.DOWN
...
+# Caches the last animation played
+var _last_animation_requested: String = "animation_not_set"
+# Caches the last animation played
+var _last_animation_played: String = "animation_not_set"
```
stickgrinder, line 116: Reading the code below, I have the impression you can avoid using two variables for last requested and last played. The value of a single variable would be updated only if the user required a different animation than the last one played, right? I may be missing something obvious here...

Also, I would not use a magic string as "null value". It is VERY unlikely, but it can collide with an actual animation. `null` should do here, since a populated string, even if empty, should not be null.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -642,29 +651,34 @@ func queue_play_animation(
...
-	# If is not present, do the same for the the fallback animation.
-	if animation == null: animation = _get_valid_oriented_animation(animation_fallback)
-	# In neither are available, exit and throw an error to check for the presence of the animations.
-	if animation == null: # Again!
-		PopochiuUtils.print_error(
-			"Neither the requested nor the fallback animation could be found for character %s.\
- Requested:%s - Fallback: %s" % [script_name, animation_label, animation_fallback]
-		)
-		return
+	if ! ((animation_label == _last_animation_requested) and (_looking_dir == _last_animation_direction)):
```
stickgrinder, line 654: Just being picky, but I'd write this as

```suggestion
	if (animation_label != _last_animation_requested) or (_looking_dir != _last_animation_direction):
```

for readability.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -742,6 +756,10 @@ func face_direction(destination: Vector2):
 	# Get the vector from the origin to the destination.
 
 	var angle = wrapf(rad_to_deg((destination - position).angle()), 0, 360)
+	var direction_faced = int(angle/45)
```
stickgrinder, line 759: You know? I'm absolutely not sure that a division, plus and integer casting, plus a variable assignment, plus an if branch, aren't slower than 16 comparisons in a worse case.

Did you benchmark it, somehow?

addons/popochiu/engine/objects/room/popochiu_room.gd
```diff
@@ -442,11 +443,15 @@ func _move_along_path(distance: float, moving_character_data: Dictionary):
 				moving_character_data.character.update_scale()
 			return
 
-		distance -= distance_between_points
-		last_point = moving_character_data.path[0]
+		# We're at the next navigation point
+		distance_to_move -= distance_to_next_navigation_point
+		last_character_position = moving_character_data.path[0]
+		if moving_character_data.path.size() > 1:
+			moving_character_data.character.face_direction(moving_character_data.path[1])
```
stickgrinder, line 450: Not sure what's the expected outcome, can you please explain?

balloonpopper: moving_character_data.path[] will have a set of coordinates where the character is to reach. e.g.
0,0 then 100,200, then 50,300, then 400, 350.
When they're at 100,200 for example, [100,200] will be path[0]. The next direction they will walk is towards path[1] i.e. [50,300], so set their facing direction towards the next leg of the path.

addons/popochiu/engine/objects/character/popochiu_character.gd
balloonpopper: I didn't benchmark it, so possibly 16 comparisons are actually faster. 
When I was talking about speed I was only referring to doing the 16 comparisons as a binary tree to limit that code section to max 4 comparisons.
This PR isn't primarily about speed (though it should be faster than the original as a lot less code is called). 
The origin of the PR was that I had to step through the code to trace something, and it was calling so many functions and evaluating so many code lines each frame that didn't need to keep getting called when values weren't changing.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: OK, makes sense. Sorry for not noticing it before!

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: OK, maybe the best option is to avoid going through a long `if/else` chain altogether.
I'm sorry for not elaborating more on this, but I have very little time at the moment, but... What if we avoid testing the angle against the values and calculate its "direction index" something like this:

```gdscript
	var angle = wrapf(rad_to_deg((destination - position).angle()), 0, 360)

	# Normalize the angle to a 0-7 range index (8 directions)
	var direction_index := int((angle + 22.5) / 45.0) % 8
	
	# Map direction index to Looking enum and animation suffixes
	match direction_index:
		0: # Right (337.5-22.5)
			_looking_dir = Looking.RIGHT
			_animation_suffixes = ['_r', '_l', '_dr', '_dl', '_d']
		1: # Down-Right (22.5-67.5) 
			_looking_dir = Looking.DOWN_RIGHT
			_animation_suffixes = ['_dr', '_dl', '_d', '_r', '_l']
		2: # Down (67.5-112.5)
			_looking_dir = Looking.DOWN
			_animation_suffixes = ['_d', '_dr', '_dl', '_r', '_l'] 
		3: # Down-Left (112.5-157.5)
			_looking_dir = Looking.DOWN_LEFT
			_animation_suffixes = ['_dl', '_dr', '_l', '_r', '_d']
		4: # Left (157.5-202.5)
			_looking_dir = Looking.LEFT
			_animation_suffixes = ['_l', '_r', '_dl', '_dr', '_d']
		5: # Up-Left (202.5-247.5)
			_looking_dir = Looking.UP_LEFT
			_animation_suffixes = ['_ul', '_ur', '_u', '_l', '_r']
		6: # Up (247.5-292.5)
			_looking_dir = Looking.UP
			_animation_suffixes = ['_u', '_ul', '_ur', '_l', '_r']
		7: # Up-Right (292.5-337.5)
			_looking_dir = Looking.UP_RIGHT
			_animation_suffixes = ['_ur', '_ul', '_r', '_l', '_u']
```

This code seems to work from a very quick test. I hope I didn't mistaken the math, but if it's ok, we can do a further step forward and get rid of the match using a map.

What I mean is, instead of "matching" the index, we can arrange the "Looking" enum so that the 0 value is RIGHT.
Also, we can have an array map for the direction suffixes, which keys are the corresponding values of the enum, and we can do:

```gdscript
func face_direction(destination: Vector2):
    var angle = wrapf(rad_to_deg((destination - position).angle()), 0, 360)
    # this requires the Looking enum to be ordered from RIGHT, clockwise
    _looking_dir := int((angle + 22.5) / 45.0) % 8
    # this requires an array or dictionary available to store the suffixes
    _animation_suffixes = _valid_animation_suffixes[_looking_dir] + ['']
```

If you want to avoid the last assignment, the _looking_dir value can be checked against a cache, but since you would have to test and reassign the _looking_dir value to the cache anyway, it's probably better to leave the `_animation_suffixes` assignment, maybe.

What do you think @balloonpopper? Makes sense?

addons/popochiu/engine/objects/room/popochiu_room.gd
stickgrinder: Imagine! I thought this was already the case... :facepalm: 
Thanks

addons/popochiu/engine/objects/character/popochiu_character.gd
balloonpopper: I'd originally thought of doing a look up table but went with my original idea because I thought that while it's fairly code efficient, it's not the most readable and easy to debug if the maths isn't quite right. I can change it to a look up table though.

stickgrinder: I don't think it's so hard to read. We can always leave comments with the calculated range and our reasoning in functions. It can feel like cluttering the code, but it makes a difference for ourselves and new contributors.

I think a table is the way to go: efficient, quick, and easy to maintain.

balloonpopper: Modified to use a table. I thought I had worked out why you added 22.5 degrees, then when looking at the code later, I couldn't remember. I tried without it and it works fine.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -642,29 +668,34 @@ func queue_play_animation(
...
+
+		# Search for a valid animation corresponding to animation_label
+		var animation = _get_valid_oriented_animation(animation_label)
+		# If is not present, do the same for the the fallback animation.
+		if animation == null: animation = _get_valid_oriented_animation(animation_fallback)
+		# In neither are available, exit and throw an error to check for the presence of the animations.
+		if animation == null: # Again!
+			PopochiuUtils.print_error(
+				"Neither the requested nor the fallback animation could be found for character %s.\
+	 Requested:%s - Fallback: %s" % [script_name, animation_label, animation_fallback]
```
stickgrinder, line 690:
```suggestion
				Requested:%s - Fallback: %s" % [script_name, animation_label, animation_fallback]
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -757,59 +782,19 @@ func face_direction(destination: Vector2):
...
+	# Get the vector from the origin to the destination.
+	var angle = wrapf(rad_to_deg((destination - position).angle()) , 0, 360)
+	# The angle calculation uses 16 angles rather than 8 for greater accuracy in choosing the facing
+	# direction animation preference.
+	var _looking_angle := int(angle / 22.5) % 16
+	_animation_suffixes = _valid_animation_suffixes[_looking_angle] + ['']
+	# The 16 directions used for animation suffixes are simplified to 8 general directions
+	_looking_dir = int(angle / 45) % 8
+	if _looking_dir == _last_looking_dir:
+		return
```
stickgrinder, line 795: This is redundant to me. Of course, you avoid an assignment, but it's idempotent, so I would keep it simple.

stickgrinder (commented):
Just a couple minor fixes and this can be merged to me!

addons/popochiu/engine/objects/character/popochiu_character.gd
balloonpopper: Removed these two lines as requested.

balloonpopper:
> Just a couple minor fixes and this can be merged to me!
Please see above comment on removing "face_direction(target_pos)" from the "walk" function.

balloonpopper:
I decided to remove the redundant face_direction call in the walk function.

mapedorr (approved):
Nice!

stickgrinder (approved):

PR #344, opened 2024-11-14
refs #343: Fix walk and look points storage and retrieval

### LINKED ISSUE #343: Can't load saved games
### Bug description

Took from the Discord server by **DiamondWarrior**:

> When I load a saved game, it has a problem with loading the character's walk to point.

image

###  Steps to reproduce

1. Run a game.
2. Save the game.
3. Try to load the game.
4. The Editor will stop at the lines shown in the screenshot.

The game should load the saved data without issues.

 - **OS type and version**: Windows 11
 - **Godot version**: 4.3
 - **Plugin version or branch**: 2.0.1

mapedorr:
Fixes #343 : The data should be stored using built-in types (int, float, String, and so on) in order to avoid hacks in Web games. A validation was added to assign the values for the `walk_to_point` and `look_at_point` based on the available data in the character's Dictionary.

### DIFF
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -241,8 +241,26 @@ func room_readied(room: PopochiuRoom) -> void:
 		chr.self_modulate = Color.from_string(chr_dic.self_modulate, Color.WHITE)
 		chr.light_mask = chr_dic.light_mask
 		chr.baseline = chr_dic.baseline
-		chr.walk_to_point = chr_dic.walk_to_point
-		chr.look_at_point = chr_dic.look_at_point
+		
+		# ---- fix #343 by checking the type of the stored value in the chr_dic Dictionary ---------
+		if chr_dic.has("walk_to_point_x"):
+			chr.walk_to_point = Vector2(chr_dic.walk_to_point_x, chr_dic.walk_to_point_y)
+		elif chr_dic.has("walk_to_point"):
+			if chr_dic.walk_to_point is String:
+				var walk_split := (chr_dic.walk_to_point as String).split_floats(",")
+				chr.walk_to_point = Vector2(walk_split[0], walk_split[1])
+			else:
+				chr.walk_to_point = chr_dic.walk_to_point
+		
+		if chr_dic.has("look_at_point_x"):
+			chr.look_at_point = Vector2(chr_dic.look_at_point_x, chr_dic.look_at_point_y)
+		elif chr_dic.has("look_at_point"):
+			if chr_dic.look_at_point is String:
+				var look_split := (chr_dic.look_at_point as String).split_floats(",")
+				chr.look_at_point = Vector2(look_split[0], look_split[1])
+			else:
+				chr.look_at_point = chr_dic.look_at_point
+		# ---------------------------------------------------------------------------- fix #343 ----
 		
 		current.add_character(chr)
 	
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -242,25 +242,11 @@ func room_readied(room: PopochiuRoom) -> void:
 		chr.light_mask = chr_dic.light_mask
 		chr.baseline = chr_dic.baseline
 		
-		# ---- fix #343 by checking the type of the stored value in the chr_dic Dictionary ---------
-		if chr_dic.has("walk_to_point_x"):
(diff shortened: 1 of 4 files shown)


addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -241,8 +241,26 @@ func room_readied(room: PopochiuRoom) -> void:
...
 		chr.baseline = chr_dic.baseline
-		chr.walk_to_point = chr_dic.walk_to_point
-		chr.look_at_point = chr_dic.look_at_point
+		
+		# ---- fix #343 by checking the type of the stored value in the chr_dic Dictionary ---------
+		if chr_dic.has("walk_to_point_x"):
+			chr.walk_to_point = Vector2(chr_dic.walk_to_point_x, chr_dic.walk_to_point_y)
+		elif chr_dic.has("walk_to_point"):
+			if chr_dic.walk_to_point is String:
+				var walk_split := (chr_dic.walk_to_point as String).split_floats(",")
```
balloonpopper, line 250: Should a check be done at this point to make sure that walk_to_point has the right format (i.e. "X,Y")? The script will die if walk_to_point contained a word for example, or didn't have a comma.

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -241,8 +241,26 @@ func room_readied(room: PopochiuRoom) -> void:
...
+				chr.walk_to_point = Vector2(walk_split[0], walk_split[1])
+			else:
+				chr.walk_to_point = chr_dic.walk_to_point
+		
+		if chr_dic.has("look_at_point_x"):
+			chr.look_at_point = Vector2(chr_dic.look_at_point_x, chr_dic.look_at_point_y)
+		elif chr_dic.has("look_at_point"):
+			if chr_dic.look_at_point is String:
+				var look_split := (chr_dic.look_at_point as String).split_floats(",")
```
balloonpopper, line 259: As above, should the format of the variable be checked? Or are you happy to assume it doesn't have an invalid value?

stickgrinder:
I would move this into a couple of functions, named something like `pack_vector_state` and `unpack_vector_state` (or anything that makes sense).

I agree with @balloonpopper about hardening the data with prior validation.

addons/popochiu/engine/interfaces/i_room.gd
mapedorr: Could be, but this assumes we're dealing with the data as it was being stored in `popochiu_room_data.gd` prior to this fix. Anyway, I can add the validation; it never hurts.

addons/popochiu/engine/interfaces/i_room.gd
mapedorr: Same reason of the previous one. I'll add the check.

mapedorr:
@balloonpopper @stickgrinder I made an update using RegEx to validate the format of the stored **String**.

Regarding the creation of a function to pack those vectors (`PopochiuCharacter.walk_to_point` and `PopochiuCharacter.look_at_point`), I’m thinking about drafting a proposal to add a feature that automatically stores any **Vector2** property in objects saved as a JSON dictionary, with `x` and `y` keys and values.

Let me know what you think, and if it’s okay to merge this PR as it is (if approved), leaving the `pack_vector_state` function for the proposal I’m referring to.

balloonpopper:
This looks complicated to me. Can we get away with just checking if it's the expected format as only Popochiu should be setting it and therefore the state should be known.
If you want to make it support all the use cases though (presumably you might be planning to use this in other places?) then it looks good.

I might be reading it wrongly (I'm not familiar enough with the codebase), but isn't the default case a dictionary? If so, I'd suggest putting the dictionary check before the regex.

mapedorr:
When you suggested including a validation, I thought of a regular expression that could also cover us in case someone manually modifies the JSON file information. 

Right now, saved games can have a string `"Vector2(x, y)"` or `"(x, y)"`, which is why I also thought a **RegEx** was a good option. Once we merge the PR, it will start saving as a **Dictionary**. I don’t mind changing the order of the checks.

stickgrinder:
@mapedorr this looks good to me. It solves the problem and is tidy.
I like your proposal for supporting Vectors state saving/loading. Can you open an "task" issue for that?

@balloonpopper do you think your open conversations could be resolved?

If you're just waiting for my validation, please go ahead and merge this fix!

stickgrinder (approved):

balloonpopper:
I'm happy that my questions have been addressed

Issue #345, opened 2024-11-15
Characters don't get deleted when changing rooms


balloonpopper:
### Bug description

Characters don't get deleted when changing rooms and appear in the next room when they haven't been included in that room.
It is related to the room not having the "Script name" property set. 

###  Steps to reproduce

Have 2 rooms where RoomA connects to RoomB. RoomA should have some characters added to it. RoomB should not have the characters added to it.
Do not set the "script_name" for the parent node of either room.
Play the game and go from RoomA to RoomB, the characters from RoomA will be instantiated in RoomB.

### Expected vs observed behavior

Characters shouldn't appear in a room they haven't been included in.

### Environment information (please complete):

- Windows 10
 - 4.4 dev 3
 - Latest popochiu dev commit 

### Additional context

After some investigation, what appears to happen is that this code runs in the goto_room function (i_room.gd): (line 166)
```
	if store_state:
		rooms_states[current.script_name] = current.state
		current.state.save_children_states()
```
If the parent node of the room does not have a script name set, then it gets the value of the destination room. This means that if you're going from RoomA -> RoomB, this store state code stores RoomA's state in the dictionary 'room_states["RoomB"]'. Afterwards, when RoomB is instantated, the code in the "room_readied" function runs and the code on line ~239-255 recreates RoomA's characters in RoomB.

```
	for chr_script_name: String in rooms_states[room.script_name]["characters"]:
		<...>
		current.add_character(chr)
```

stickgrinder:
@balloonpopper sorry for the lag.

I'm going to close this without fixing because we can't accept bugs on a non-supported version of Godot.
In the past we notices that working with dev engine versions shown bugs that eventually disappeared at a later stage of Godot development.

Thanks for taking the time to report, and work this around but we'll face the problem when we'll port the engine to 4.4 stable.

balloonpopper:
@stickgrinder  - I disagree with the closing of this ticket. Why didn't you just ask if I could confirm it in 4.3 instead of closing it?

This should be reopened. I have just tested this in 4.3 and this is **still a valid bug**. When travelling between 2 rooms without script names associated with them, characters from the first room appear in the second room when they shouldn't.

In fact, it's more than just characters. all these errors appear when I go from the first room to the second room. They are all things in the first room that should get deleted when entering the second room.
![image](https://github.com/user-attachments/assets/128b5225-fe75-45c6-8840-686033928d5f)

stickgrinder:
[...]

PR #346, opened 2024-11-15
issue_345 Set a room script name if one wasn't set

balloonpopper:
Fixes https://github.com/carenalgas/popochiu/issues/345

I'm assuming the state will always have the name based off what I saw in testing. If this isn't a valid assumption, maybe it would be better to crash the game with an error message if the script name for a room is not set.

### DIFF
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -198,6 +198,10 @@ func room_readied(room: PopochiuRoom) -> void:
 	if not is_instance_valid(current):
 		current = room
 	
+	# Apply a room script_name if it hasn't been explicitly set for the room
+	if room.script_name == "":
+		room.script_name = room.state.script_name
+
 	# When running from the Editor the first time, use goto_room
 	if Engine.get_process_frames() == 0:
 		await get_tree().process_frame


balloonpopper:
The current code doesn't solve all the problems. The "tool" scripts have the buttons for creating props/hotspots etc all greyed out if a scene without a script name is loaded. The only way to fix it is to give the scene a script name, then close and reopen the scene. Not sure how to fix this.

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -198,6 +198,10 @@ func room_readied(room: PopochiuRoom) -> void:
 	if not is_instance_valid(current):
 		current = room
 	
+	# Apply a room script_name if it hasn't been explicitly set for the room
+	if room.script_name == "":
+		room.script_name = room.state.script_name
```
mapedorr, line 203: Maybe we could check if `room.state.script_name` is not empty. In case it is, we could assume the `script_name` for the room as the name of its `.tscn` file (`room.scene_file_path`) without the `room_` preffix.

balloonpopper: Thinking about this further, I think the script name should be changed (I'm not sure how at this point) by the tool part of the code rather than at runtime. I have this sort of code in my scripts for setting the player's start position in a room:

```
if C.player.last_room == R.LockerRoom.script_name:
```
If in my room (i.e. "LockerRoom") the script name is not set, then I expect this test to fail. I haven't tested this yet though.

stickgrinder:
@balloonpopper after talking to @mapedorr, and considering that #345 can't be accepted, we are going to close this PR.

We prefer to wait until 4.4 is in a more stable state, and - if the problem is still there - fix the behavior the proper way.
Your workaround does the trick but it adds a bit of tech-debt.

Thanks for taking the time.

PR #358, opened 2024-11-24
refs #348: Show the TransitionLayer when no using transition animation

### LINKED ISSUE #348: Room contents is playing before  _on_room_entered()  has executed when no transition is enabled
### Bug description

This function is supposed to run when "the room is in the tree but it is not visible".
If you change room with no transition (e.g. "R.goto_room("Locker", false)") then you can see the contents of the new room ("Locker") briefly while its _on_room_entered() function executes. I can tell this as I have calls in _on_room_entered() to disable various things
e.g. 
```
	if C.player.state.WearingTowel == false:
		R.get_prop("Towel").show()
		R.get_prop("Uniform").hide()
```
The Uniform prop displays for a brief period before disappearing.

###  Steps to reproduce

1. Create a room with a prop
2. In its _on_room_entered() function, hide the prop
3. Create a second room, with a "goto_room" function in it that takes you to the first room without a transition
4. Play the game. Observe when you move from the second room to the first room that the prop displays temporarily before disappearing

Everything in _on_room_entered should complete before the room is drawn to the screen

 - Windows 10
 - 4.3 dev 3
 - Latest development branch of Popo

mapedorr:
Fixes #348 : Rooms should be in black while their _on_room_entered() function is executing no matter if devs are using or not transition animations. This allows to execute logic that can affect the rendering of the room without allowing players to see what's happening.

### DIFF
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -154,6 +154,8 @@ func goto_room(
 	if use_transition and Engine.get_process_frames() > 0:
 		E.tl.play_transition(E.tl.FADE_IN)
 		await E.tl.transition_finished
+	elif Engine.get_process_frames() > 0:
+		E.tl.show_curtain()
 	
 	# Prevent the GUI from showing info coming from the previous room
 	G.show_hover_text()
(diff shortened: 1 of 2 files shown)


mapedorr:
@balloonpopper Could you test this so we know if it makes the flow behave as you expect?

stickgrinder:
LGTM, just in case you wonder :smile:

balloonpopper:
The scenes that I was seeing the problem with are working correctly - thanks @mapedorr

Issue #360, opened 2024-11-29
The dialog resize function does not work in the 2 button interface


balloonpopper:
### Bug description

The _modify_size in dialog_caption.gd doesn't modify the size of the text box correctly -  it sets the height of the RichTextLabel, but not the width. Short text lines therefore appear at a different y coordinate to long ones.

Additionally, as text is being typed by the tween, if there's not enough room on the current line for a word, the tween writes whatever part of the word it can fit on the line then erases the half word and rewrites it on the following line before completing it.

###  Steps to reproduce

In a 1080p resolution game, use commands like this
```
await C.get_OfficerBlonde().say("Hi!")
await C.get_OfficerBlonde().say("Do you know the best thing about this shower, Sonny?")

```
to write text to the screen. According to dialog_text.gd, "wrap_width := 200.0" but the long line will be displayed ignoring that it goes over 200px.
Also, you'll see that the y coordinate of the two lines is very different.

### Expected vs observed behavior

Width enforcement should be respected.
Words should not be half written if they don't fit on a line but should be entirely written in their final form on a brand new line.

### Environment information (please complete):

 - Windows 10
 - 4.3
 - Latest dev branch

stickgrinder:
@mapedorr can't remember if we actually fixed that. Let me know so I can close this.
Thanks!

PR #361, opened 2024-11-29
refs #360: Fix dialog_caption size

balloonpopper:
Fixes https://github.com/carenalgas/popochiu/issues/360

### DIFF
--- a/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
+++ b/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
@@ -2,7 +2,7 @@ extends PopochiuDialogText
 
 
 #region Private ####################################################################################
-func _modify_size(msg: String, target_position: Vector2) -> void:
+func _modify_size(msg: String, _target_position: Vector2) -> void:
 	var _size := await _calculate_size(msg)
 
 	# Define size and position (before calculating overflow)
@@ -12,31 +12,32 @@ func _modify_size(msg: String, target_position: Vector2) -> void:
 	rich_text_label.position.y = get_meta(DFLT_POSITION).y - (_size.y - get_meta(DFLT_SIZE).y)
 
 
+func _append_text(msg: String, props: Dictionary) -> void:
+	msg = _correct_line_breaks(msg)
+	rich_text_label.text = "[center][color=%s]%s[/color][/center]" % [props.color.to_html(), msg]
+
+
 ## Appends text for the dialog caption
 ## Ensures that where a printing a word would see it wrap to the next line that the newline
 ## is forced into the text. Without this the tween in dialog_text.gd would print part of the word
-## until it runs out of space, then erase the part word and put it onto the next line which looks
+## until it runs out of space, then erase the part word and rewrite it on the next line which looks
 ## messy.
-func _append_text(msg: String, props: Dictionary) -> void:
+func _correct_line_breaks(msg: String) -> String:
 	rich_text_label.text = msg
 	var number_of_lines_of_text := rich_text_label.get_line_count()
 	if number_of_lines_of_text > 1:
 		var current_line_number := 0
 		for current_character in range(0, rich_text_label.text.length()):
-
-			var ThisChar = rich_text_label.text[current_character]
-			var ThisLine = rich_text_label.get_character_line(current_character)
 			if rich_text_label.get_character_line(current_character) > current_line_number:
 				current_line_number += 1
 				if rich_text_label.text[current_character-1] == " ":
 					rich_text_label.text[current_character-1] = "\n"
 				elif rich_text_label.text[current_character-1] != "\n":
-					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" + rich_text_label.text.right(-current_character)
+					rich_text_label.text = rich_text_label.text.left(current_character) +\
+					"\n" + rich_text_label.text.right(-current_character)
 
 				if current_line_number == number_of_lines_of_text - 1:
 					break
-
-	msg = rich_text_label.text
-	rich_text_label.text = "[center][color=%s]%s[/color][/center]" % [props.color.to_html(), msg]
+	return rich_text_label.text
 
 #endregion
--- a/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_overhead/dialog_overhead.gd
+++ b/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_overhead/dialog_overhead.gd
@@ -54,15 +54,13 @@ func _correct_line_breaks(msg: String) -> String:
 	if number_of_lines_of_text > 1:
 		var current_line_number := 0
 		for current_character in range(0, rich_text_label.text.length()):
-
-			var ThisChar = rich_text_label.text[current_character]
-			var ThisLine = rich_text_label.get_character_line(current_character)
 			if rich_text_label.get_character_line(current_character) > current_line_number:
 				current_line_number += 1
 				if rich_text_label.text[current_character-1] == " ":
 					rich_text_label.text[current_character-1] = "\n"
 				elif rich_text_label.text[current_character-1] != "\n":
-					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" + rich_text_label.text.right(-current_character)
+					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" +\
+					rich_text_label.text.right(-current_character)
 
 				if current_line_number == number_of_lines_of_text - 1:
 					break
--- a/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
+++ b/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
@@ -58,15 +58,13 @@ func _correct_line_breaks(msg: String) -> String:
 	if number_of_lines_of_text > 1:
 		var current_line_number := 0
 		for current_character in range(0, rich_text_label.text.length()):
-
-			var ThisChar = rich_text_label.text[current_character]
-			var ThisLine = rich_text_label.get_character_line(current_character)
 			if rich_text_label.get_character_line(current_character) > current_line_number:
 				current_line_number += 1
 				if rich_text_label.text[current_character-1] == " ":
 					rich_text_label.text[current_character-1] = "\n"
 				elif rich_text_label.text[current_character-1] != "\n":
-					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" + rich_text_label.text.right(-current_character)
+					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" +\
+					rich_text_label.text.right(-current_character)
 
 				if current_line_number == number_of_lines_of_text - 1:
 					break


addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
```diff
@@ -2,16 +2,41 @@ extends PopochiuDialogText
...
+		for current_character in range(0, rich_text_label.text.length()):
+
+			var ThisChar = rich_text_label.text[current_character]
+			var ThisLine = rich_text_label.get_character_line(current_character)
+			if rich_text_label.get_character_line(current_character) > current_line_number:
+				current_line_number += 1
+				if rich_text_label.text[current_character-1] == " ":
+					rich_text_label.text[current_character-1] = "\n"
+				elif rich_text_label.text[current_character-1] != "\n":
+					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" + rich_text_label.text.right(-current_character)
```
mapedorr, line 34: This line is too long. Could you split it so it doesn't go past column 100?

mapedorr (changes requested):
I'm asking for a couple of changes. I still have to devote some time to test the behavior of the component.

balloonpopper:
The changes to the _calculate_size function were because the existing code would split the text from this command onto 2 lines for a 100px text box, when the text was only about 60px wide when tested with (from memory) dialog_overhead.
```
await C.get_OfficerBlonde().say("It's free.")
```

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
balloonpopper: Fixed

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_text.gd
```diff
@@ -79,7 +79,6 @@ func play_text(props: Dictionary) -> void:
 	await _modify_size(msg, props.position)
 	
 	# Assign the text and align mode
-	msg = "[color=%s]%s[/color]" % [props.color.to_html(), msg]
```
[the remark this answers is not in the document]
balloonpopper: Duplicated to the overhead and portrait functions.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
```diff
@@ -4,14 +4,40 @@ extends PopochiuDialogText
 #region Private ####################################################################################
 func _modify_size(msg: String, _target_position: Vector2) -> void:
 	var _size := await _calculate_size(msg)
-	
+
 	# Define size and position (before calculating overflow)
+	rich_text_label.size.x = _size.x
 	rich_text_label.size.y = _size.y
+	rich_text_label.position.x = (get_viewport_rect().size.x/2) - (_size.x /2)
```
mapedorr, line 11: I think the ideal approach in this case is not to modify the width (`size.x`) and X position of the **RichTextLabel**. Considering how this **Control** is configured in Godot, what we should do is ensure that the value defined in the `wrap_width` property is assigned to the `size.x` property of the **RichTextLabel** in the `_ready()` function. To control the Y position of the component, Godot's anchors can be used.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
```diff
@@ -4,14 +4,40 @@ extends PopochiuDialogText
...
+				current_line_number += 1
+				if rich_text_label.text[current_character-1] == " ":
+					rich_text_label.text[current_character-1] = "\n"
+				elif rich_text_label.text[current_character-1] != "\n":
+					rich_text_label.text = rich_text_label.text.left(current_character) +\
+					"\n" + rich_text_label.text.right(-current_character)
+
+				if current_line_number == number_of_lines_of_text - 1:
+					break
+	return rich_text_label.text
```
mapedorr, line 41: I like what this code does, but it shouldn't be duplicated in the 3 **DialogText** components. You should move this to the `dialog_text.gd` script so any **DialogText** component can make use of it.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_overhead/dialog_overhead.gd
```diff
@@ -44,4 +43,28 @@ func _get_icon_to_position() -> float:
...
+					rich_text_label.text[current_character-1] = "\n"
+				elif rich_text_label.text[current_character-1] != "\n":
+					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" +\
+					rich_text_label.text.right(-current_character)
+
+				if current_line_number == number_of_lines_of_text - 1:
+					break
+	return rich_text_label.text
+
+
```
mapedorr, line 69: Remove this after moving this function to the `dialog_text.gd` script.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_overhead/dialog_overhead.gd
```diff
@@ -27,13 +26,13 @@ func _set_default_label_size(lbl: Label) -> void:
 
 
 func _append_text(msg: String, props: Dictionary) -> void:
+	msg = _correct_line_breaks(msg)
+
```
mapedorr, line 30: Assign the `[color][/color]` to the `msg` in the same way you did for the `dialog_caption.gd:17` and the `dialog_portrait.gd:47`.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_portrait/dialog_portrait.gd
```diff
@@ -42,4 +42,32 @@ func _set_default_size() -> void:
...
+				if rich_text_label.text[current_character-1] == " ":
+					rich_text_label.text[current_character-1] = "\n"
+				elif rich_text_label.text[current_character-1] != "\n":
+					rich_text_label.text = rich_text_label.text.left(current_character) + "\n" +\
+					rich_text_label.text.right(-current_character)
+
+				if current_line_number == number_of_lines_of_text - 1:
+					break
+	return rich_text_label.text
+
```
mapedorr, line 72: Remove this after moving this function to the `dialog_text.gd` script.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_text.gd
```diff
@@ -73,14 +73,18 @@ func play_text(props: Dictionary) -> void:
 	
 	if PopochiuConfig.should_talk_gibberish():
 		msg = D.create_gibberish(msg)
-	
+
 	# Call the virtual method that modifies the size of the RichTextLabel in case the dialog style
 	# requires it.
 	await _modify_size(msg, props.position)
 	
+	rich_text_label.push_color(props.color)
```
mapedorr, line 81: We removed this in #357 because it was causing issues when centering text. Please remove it again since each **DialogText** component should assign the color itself using BBCode (`[color][/color]`).

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_text.gd
```diff
@@ -73,14 +73,18 @@ func play_text(props: Dictionary) -> void:
...
+	
 	# Assign the text and align mode
-	msg = "[color=%s]%s[/color]" % [props.color.to_html(), msg]
-	_append_text(msg, props)
+	if PopochiuConfig.should_talk_gibberish():
+		_append_text(D.create_gibberish(msg), props)
+	else:
+		_append_text(msg, props)
```
mapedorr, line 87: This should be just this since the gibberish text replacement was already done in line 75.

```suggestion
	_append_text(msg, props)
```

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_text.gd
```diff
@@ -169,46 +173,37 @@ func _modify_size(_msg: String, _target_position: Vector2) -> void:
...
 	rich_text_label.add_child(rt)
 	
 	# Create a Label to check if the text exceeds the wrap_width
 	var lbl := Label.new()
 	lbl.add_theme_font_override("normal_font", get_theme_font("normal_font"))
 	
-	_set_default_label_size(lbl)
-	
 	lbl.text = rt.get_parsed_text()
+	lbl.size = lbl.get_minimum_size()
```
mapedorr, line 191: I think this is not necessary since the `custom_minimum_size` of the component at this point will be **Vector2.ZERO**.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_text.gd
```diff
@@ -169,46 +173,37 @@ func _modify_size(_msg: String, _target_position: Vector2) -> void:
 
 
 ## Creates a RichTextLabel to calculate the resulting size of this node once the whole text is shown.
+## Uses a RichTextLabel to provide the "get_parsed_text" function, and a label within it to work out
+## the minimum size. Calculating the size from just the RichTextLabel does not work.
```
mapedorr, line 177: If you want to replace the description of the function, please remove the previous one. Also, I suggest a small change to make it clear what the function really does.

```suggestion
## Uses a [RichTextLabel] and a [Label] to work out the minimum size since calculating it from just
## the [RichTextLabel] does not work.
```

mapedorr (changes requested):
I request some changes before merging this to `develop`.

mapedorr:
I think the `_correct_line_breaks()` function you created to modify the text so the line breaks are displayed properly works fine, and it actually improves the behavior of the tween, but I suggest making that calculation before trying to calculate the size of each **DialogText** component.

## About the Overhead dialog
The idea of this component is to take into account the `dialog_pos` property of each **PopochiuCharacter** to make the bottom of the component to render above that line. In many cases, your implementation makes the text to render below that line, which can cause reading issues on long texts.

👇 **Your implementation** 👇

https://github.com/user-attachments/assets/9c07f214-2ffe-4f89-a4f7-f21769b62a98


👇 **The current (an expected) behavior** 👇

https://github.com/user-attachments/assets/bd3df226-ca60-444b-a471-ce1a74989a5a

## About the Portrait dialog
Due to how this scene is structured, the size of the **RichTextLabel** adjusts to the available space within the **PanelContainer** component where it is contained. Perhaps what should happen with this component is that the value of the `wrap_width` property defines the size that the **PanelContainer** should have. In any case, modifying the text using your `_correct_line_breaks()` function can help the make the tween look better.

## About the Caption dialog
As I mentioned in one of the comments, I think the ideal approach for this component is that the value of the `wrap_width` property defines the X size of the **RichTextLabel** component on its `_ready()` function. The _width_ of this component will remain fixed regardless of the amount of content to render. I think modifying only the height based on the content is fine (as it is in the current implementation (in _develop_)), but we could use the `_correct_line_breaks()` function to ensure the final size adjusts better based on the final text content.

# To summarize
I think the right approach to make use of the code you made for calculating the line breaks should be used before calling `_modify_size()` (line 79 in your code) in the **DialogText** class. That way, the calculation of the size of the **RichTextLabel** component will use the text with the line breaks fixes.

addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
```diff
@@ -4,14 +4,40 @@ extends PopochiuDialogText
...
+	msg = _correct_line_breaks(msg)
 	rich_text_label.text = "[center][color=%s]%s[/color][/center]" % [props.color.to_html(), msg]
 
 
+## Appends text for the dialog caption
+## Ensures that where a printing a word would see it wrap to the next line that the newline
+## is forced into the text. Without this the tween in dialog_text.gd would print part of the word
+## until it runs out of space, then erase the part word and rewrite it on the next line which looks
+## messy.
+func _correct_line_breaks(msg: String) -> String:
```
mapedorr, line 25: I think `_fix_line_breaks` is a better name for this function. "Correct" sounds weird.

stickgrinder:
Reviewing this with @mapedorr we agree this component may needs improvement, but the proposed implementation has some downsize.

First of all, forcefully changing the actual string may break the localization (when it will be available), since the string as it appears in the script is the key for the translation.

Also there is a lot of code which is probably better handled by the nodes internals (we should be able to rely on the standard behavior of RichTextLabels for most of the solution).

We'll keep the issue open for review, but are closing this PR.

PR #362, opened 2024-12-04
Replace uses of Global classes inside plugin scripts

### LINKED ISSUE #250: When installing popochiu 2.0 for godot 4.2.2 multiple non-fatal errors occure
### Bug description

grafik

###  Steps to reproduce

1. Download [Popochiu 2.0 - Beta 3](https://github.com/carenalgas/popochiu/releases/download/v2.0.0-beta3/popochiu-v2.0.0-beta3.zip)github.com/carenalgas/popochiu/releases/download/v1.8.7/popochiu-v1.8.7.zip).
2. Extract it and copy the addons folder into your project folder.

No errors.

 - **OS type and version**: Win10
 - **Godot version**: 4.2.2
 - **Plugin version or branch**: popochiu 2.0

mapedorr:
Fixes #250 : To prevent Godot from displaying a series of error messages when the plugin is added to a project, this PR removes direct calls to its Global classes (**E**, **R**, **C**, **I**, **D**, **A**, **G**, **Cursor**, and **Globals**) across all classes on the plugin side. Instead, one must use the `static` variables created in **PopochiuUtils**: (**e**, **r**, **c**, **i**, **d**, **a**, **g**, **cursor**, and **globals**).  

This change also makes it possible to start using the plugin without restarting Godot after enabling it.


addons/popochiu/editor/factories/factory_base_popochiu_obj.gd
```diff
@@ -149,11 +149,11 @@ func _create_script_from_template() -> int:
 		"%s_state_template" % _type_label,
 		"%s_%s_state" % [_type_label, _snake_name]
 	)
-
 	new_code = new_code.replace(
 		"Data = null",
 		'Data = load("%s.tres")' % _path_base
 	)
+	new_code = new_code.replace("PopochiuUtils.e", "E")
```
stickgrinder, line 156: clever!

addons/popochiu/engine/interfaces/i_audio.gd
```diff
@@ -22,6 +22,13 @@ extends Node
 var twelfth_root_of_two := pow(2, (1.0 / 12))
 
 
+#region Godot ######################################################################################
+func _init() -> void:
+	Engine.register_singleton(&"A", self)
```
stickgrinder, line 27: Interesting! Mind explaining why are you doing it here?

addons/popochiu/engine/objects/gui/components/sound_volumes/sound_volumes.gd
```diff
@@ -63,9 +62,10 @@ func _on_audio_manager_ready() -> void:
 		$SlidersContainer.add_child(slider)
 		
 		slider.value_changed.connect((
-			func (value: float, bus_name: String):
-				E.am.set_bus_volume_db(
-					bus_name,
+			func (value: float, bus_name_param: String):
+				@warning_ignore("standalone_ternary")
```
stickgrinder, line 66: Darn, I'm torn. Passing a value as a parameter like this is very punk and very smart at the same time...

stickgrinder (approved):
Out of curiosity, I have a question about the init functions that now register the singletons. Where were you doing it before?

Anyway, this can be merged for me!

addons/popochiu/engine/objects/gui/components/sound_volumes/sound_volumes.gd
balloonpopper: I'm going to say that I have no idea what this piece of code is doing. Is it possible to simplify this by any chance?

stickgrinder: I tend to agree with @balloonpopper on that. It's nothing bad, and it's something pretty readable to (ex.) a skilled javascript developer. I can follow along (despite not being a skilled js dev myself), bit the smell here is the need to suppress a warning. It's not wrong, it's stretching the language beyond its paradigm.

addons/popochiu/engine/interfaces/i_audio.gd
mapedorr: Because this is the function called by Godot when an object is initialized. That way we assure the first thing the object does is to register itself as a Singleton. To be honest, a saw this pattern in another plugin and I liked it. Before this PR we were not registering our Global classes as singletons.

addons/popochiu/engine/objects/gui/components/sound_volumes/sound_volumes.gd
mapedorr: Thanks. I put the logic in a separate function now.

PR #363, opened 2024-12-13
Fix double call to window close which causes save to run twice

balloonpopper:
Close is called from the end of _on_ok  which doubles up with the call to close in popochiu_popup.gd  in the on_ok_pressed function. This causes the save function to run a second time unnecessarily.


balloonpopper:
Code in  on_ok_pressed : 
```
func on_ok_pressed() -> void:
	_on_ok()
	close()
```

stickgrinder:
@balloonpopper thanks for that, nice spot!

PR #364, opened 2024-12-14
WIP : Add screenshots to save games

balloonpopper:
Proposal to add screenshots to save games. It works at the moment, but I'd like the project maintainers to review the code for feedback.

I'm not sure how much the save game interface should be viewed as an example for people wanting to expand on it, and how much it should be modified to make it more how I envisage most people would want it to behave. (I appreciate that how I think people would want it to behave isn't necessarily correct - but it does look very basic right now and I think it would be better if it was more fully featured). I propose the following changes (but the current code may be close enough to "done" if others don't agree with my suggestions as the interface should be looked at as an example only : 

- Date/time isn't allocated to the save game file until after you hit ok. Currently clicking a save game slot and having it not actually save until you've hit ok - even though it's put a timestamp on the slot - is misleading. I would suggest modifying this behaviour.

- Expanding on the above idea, clicking a slot should either immediately save if the slot is empty, or ask for confirmation on an overwrite. I would remove the ok button entirely.

- I'd like to see the save game you're hovering over highlighted by drawing a box around it / potentially dimming all other save game slots.

- The interface gets very buggy if I don't pre-resize the screenshot prior to trying to show it on the load/save screen. The resolution of the resized image, and the filename to save games to should be exported to project settings. Currently it's hard coded in the file but I'll hold off on moving it until the project maintainers advise if at the top of the file or in project settings would be preferred.


balloonpopper:
Changed "New Save" to be "Empty Slot", and also added a "Missing Image" image in case you have a save game file but not its related image.

balloonpopper:
I tried doing further updates to the interface (like drawing a box around the save game entry on mouse-over) but I gave up as I found it too inconsistent with the current popochiu_popup class, and Godot's control nodes are painful to use.

stickgrinder:
Hi @balloonpopper.

Me and @mapedorr just reviewed your implementation. It's working code but there are some corner cases that makes a proper solution pretty hard to complete.

To start with, this implementations should not overwrite the save/load component we have in place. A new one should be created instead.
Saving screenshot in a file may be a problem with web builds of the games, and this have to be checked better.
Also, your comment about Godot controls being a bit finnicky is not far from truth.

Considering all of that, the best option may be for @mapedorr to implement this feature. He has thorough knowledge of Godot controls, plus he knows the GUI system ins and outs.

I think we can work this out under the umbrella of #365 if we want to broaden its scope.
Are you ok with this?

balloonpopper:
If @mapedorr wants to implement this feature then that's fine.
You mentioned above that you don't want to modify the existing load/save functionality. In order to implement that I assume that you therefore would set up something similar to the GUI selection so you can select the save file type that you like? I have concerns about using the same setup tool for this - if I want to change my save game style, I don't want to overwrite my GUI type in case I have code customisations in there. The setup tool would need to be overhauled so that it only impacts the components you want to change.

stickgrinder:
> The setup tool would need to be overhauled so that it only impacts the components you want to change.

Yes, there is draft work by @mapedorr on that that we decided to remove from 2.0, but will make its way into future versions, to allow people to pick up components _à-la carte_. The plugin UI to "compose" your game UI was too green and still requires proper design.

In the meantime, you are right: customizations need to be managed by hand. But this is true for every GUI component that gets customized. The best option we have right now is writing more documentation on how to incorporate new or updated components into the game UI if you upgrade Popochiu in your project.

Please, note that there has been a lot of discussion on how to support devs creating game UIs. It's not an easy deed, because GUIs are the most prominent elements that change a lot between adventure games (much more than characters, props, rooms or dialogues, which behaviors are 90% the same across different games).

Contributions in the GUI area will probably need a good amount of upfront design to be considered for inclusion. If you want to take this over, of course, the best option can be to detail the solution architecture in an issue before getting to the implementation.

stickgrinder:
We reworked #365 to broaden the scope, integrating a sshot-based save functionality.

@mapedorr will take this over from scratch because many things changed in the GUIs in the last months and this implementation would not be easy to integrate.

PR #367, opened 2024-12-23
Pause gameplay while popups are visible

balloonpopper:
Adds pausing functionality for when popups appear

Unpausing is done in each popup individually as I didn't want the animations in the room to be running until the fade completed, so the restart needed to be in each popup rather than a single call in popochiu_popup.gd

Known bugs that I haven't been able to fix (some help would be appreciated)
1) Character facing direction isn't corrected until after the game is unpaused. The fixes I've tried haven't worked.
2) If a character is walking when you save the game  - e.g. 
```
			await C.get_OfficerBrown().walk_to_marker("ExitPoint")
```
the character's walking state is not restored when you load the save. This appears to be a generic save-game bug, and not one in my code.

### DIFF
--- a/addons/popochiu/engine/cursor/cursor.gd
+++ b/addons/popochiu/engine/cursor/cursor.gd
@@ -29,6 +29,8 @@ var is_blocked := false
 #region Godot ######################################################################################
 func _init() -> void:
 	Engine.register_singleton(&"Cursor", self)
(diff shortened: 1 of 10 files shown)


stickgrinder:
I created #455 to better manage this need. Closing this PR for now.

Issue #245, opened 2024-07-06
Add a dropdown to Props' Inspector to define the `link_to_item` property


mapedorr:
### Benefit description

Instead of requiring developers to write the `script_name` of the inventory item they want to link a prop to (which is more prone to typos), it would be good if the property in the **Inspector** was a dropdown menu with the game's inventory items as options.

### Solution description

An **EditorInspectorPlugin** can be created (`prop_inspector_plugin.gd`) to change the **Control** node used to render the `link_to_item` property to an **OptionButton** whose items are the inventory items in the game. This menu would be empty by default, and would show the option that matches the value of the `link_to_item` property (if any) of each prop.

### Exclusions

None.

### Implications

This won't require doing anything related to the Migrations tool.

PR #380, opened 2025-01-10
Improve UX by showing a dropdown of inventory items for prop's link_to_item property.

mapedorr:
Implements #245 to avoid mistakes or typos when linking an inventory item to a prop so it hides when the linked item is in the inventory.

### DIFF
--- a/addons/popochiu/editor/inspector/prop_inspector_plugin.gd
+++ b/addons/popochiu/editor/inspector/prop_inspector_plugin.gd
@@ -1,0 +1,63 @@
+extends EditorInspectorPlugin
+
+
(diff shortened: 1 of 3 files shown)


stickgrinder:
@mapedorr LGTM.

The only doubt is what happens in a game with a lot of inventory items. If the options button allows to search by typing (like jumping to the letter pressed) I see no problem: go ahead and merge it.

If not, I would improve this by adding a filter on top of the list *if it's trivial*. Should it require significative effort, ignore this comment and merge.

stickgrinder (approved):

mapedorr:
Ok. I'll if I can make the change quickly.

Issue #381, opened 2025-01-15
Error when character animation is missing.


cidneyhamilton:
### Bug description

If a character is missing an animation (placeholder art without a separate walk or idle), etc, popochiu_character.gd will throw an exception (_get_valid_oriented_animation will return null when it expects a string-- default value is "null".
###  Steps to reproduce

1. Create a new main character.
2. Import from an aseprite template with a missing walk animation.
3. Add character to scene, attempt to play.

### Expected vs observed behavior

Should fallback sensibly (default to idle if walk is missing), as per previous version.

PR #382, opened 2025-01-15
Fix bug with missing character animation.

cidneyhamilton:
This fixes https://github.com/carenalgas/popochiu/issues/381 by returning the default for _current_animation (the string "null") instead of null.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -683,9 +683,10 @@ func play_animation(animation_label: String, animation_fallback := 'idle'):
 		# Search for a valid animation corresponding to animation_label
 		_current_animation = _get_valid_oriented_animation(animation_label)
 		# If is not present, do the same for the the fallback animation.
-		if _current_animation == "null": _current_animation = _get_valid_oriented_animation(animation_fallback)
+		if _current_animation.is_empty():
+			_current_animation = _get_valid_oriented_animation(animation_fallback)
 		# In neither are available, exit and throw an error to check for the presence of the animations.
-		if _current_animation == "null": # Again!
+		if _current_animation.is_empty(): # Again!
 			PopochiuUtils.print_error(
 				"Neither the requested nor the fallback animation could be found for character %s.\
 				Requested:%s - Fallback: %s" % [script_name, animation_label, animation_fallback]
@@ -933,7 +934,7 @@ func _get_valid_oriented_animation(animation_label):
 		if animation_player.has_animation(animation):
 			return animation
 
-	return "null"
+	return ""
 
 
 func _walk_to_node(node: Node2D, offset: Vector2) -> void:
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -43,6 +43,9 @@ signal move_ended
 ## Emitted when the animation to grab things has finished.
 signal grab_done
 
+## Empty string constant to perform type checks (String is not nullable in GDScript. See #381, #382).
+const EMPTY_STRING = ""
+
 ## The [Color] in which the dialogue lines of the character are rendered.
 @export var text_color := Color.WHITE
 ## Depending on its value, the [b]$Sprite2D[/b] child will be flipped horizontally depending on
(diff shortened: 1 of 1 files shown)


stickgrinder:
Hi @cidneyhamilton! Sorry for taking so long with this review.

This error stems from merging #341 that from a certain point of view did more harm than good, in retrospective.
Now, your solution is coherent with how `_current_animation` is defined at `popochiu_character.gd:112` but that `"null"` string slipped through my original review and I don't like it at all.

I'm for strong typing but using strings like that is a terrible idea (what if someone use "null" as the valid name for an animation? That's out of our control, so we need to stick with `null` value.

The problem is [GDScript doesn't support nullable types](https://github.com/godotengine/godot-proposals/issues/162) ([yet](https://github.com/godotengine/godot/pull/76843)) so the only valid option left is to relax the typing of the following character's private properties:

* `_current_animation`
* `_last_requested_animation_label`
* `_last_requested_animation_dir`

Yes, I would not like my mom to know and yes, I love strong typing, but since GDScript still lacks features that would be sensible in a dynamic language, we're left with untyped vars for this purpose.

So the plan is to revert your commit and make these properties untyped.
Let me know if you want to do it or if you prefer me pass the torch to me. Thanks for your help.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -683,9 +683,9 @@ func play_animation(animation_label: String, animation_fallback := 'idle'):
 		# Search for a valid animation corresponding to animation_label
 		_current_animation = _get_valid_oriented_animation(animation_label)
 		# If is not present, do the same for the the fallback animation.
-		if _current_animation == null: _current_animation = _get_valid_oriented_animation(animation_fallback)
+		if _current_animation == "null": _current_animation = _get_valid_oriented_animation(animation_fallback)
```
mapedorr, line 686: Instead of asking for `"null"`, I think it will be better if we ask for `is_empty()`.

```suggestion
		if _current_animation.is_empty():
			_current_animation = _get_valid_oriented_animation(animation_fallback)
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -683,9 +683,9 @@ func play_animation(animation_label: String, animation_fallback := 'idle'):
...
 		# In neither are available, exit and throw an error to check for the presence of the animations.
-		if _current_animation == null: # Again!
+		if _current_animation == "null": # Again!
```
mapedorr, line 688: Same here.

```suggestion
		if _current_animation.is_empty(): # Again!
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -933,7 +933,7 @@ func _get_valid_oriented_animation(animation_label):
 		if animation_player.has_animation(animation):
 			return animation
 
-	return null
+	return "null"
```
mapedorr, line 936:
```suggestion
	return ""
```

Additionally, the signature of the method should also be updated to include the returning type: `func _get_valid_oriented_animation(animation_label: String) -> String:`

mapedorr (changes requested):
I'll complement this with an additional comment on the PR.

mapedorr:
I agree with @stickgrinder that we shouldn't validate against `"null"`. We missed this detail when we merged the changes from #341, oops!  

We can keep the `_current_animation` and `_last_requested_animation_label` variables as **String**, but instead of setting them to `"null"` in their definition (lines 112 and 114, respectively), they should default to empty strings `""`.  

I propose that instead of validating against `"null"` as suggested in this PR, we use the `is_empty()` method to check whether the string is empty or not. This also means that the `_get_valid_oriented_animation()` method should return `""` instead of `null`.

Sorry for letting so much time pass before reviewing this PR. 😣

stickgrinder:
@mapedorr I accepted your changed and introduced a strong-typed constant for type-checking.
The main reason is, we're using empty strings as null values here and having a constant makes it less error-prone (no way you can type a space between brackets and introduce silent-regressions)

If you deem it overkill, I'll revert my last commit. If not, we can merge this fix.

stickgrinder (approved):

mapedorr (approved):
Woooo. I like this. That `const EMPTY_STRING` make it look more professional 🤓 🤓 🤓

Issue #383, opened 2025-01-16
Documentation: restructure the main navigation and reposition some sections.


stickgrinder:
### Benefit description

We stubbed a structure of the documentation before 2.0 release, and it made sense on paper.

At this point thou, some sections outgrewed their initial scopes (mainly, the "Create a game stub" section, who became so large it can actually be considered as the content of the "How to develop a game" section, maybe with some additions?

So, we are going to reorganize the navigation, removing redundancy and making the most out of what we already have, so we can focus on the missing parts.

### Solution description

* [x] #387
* [x] Add and "Advanced techniques" or  "Guides and how-tos" section for advanced or more vertical stuff
* [x] #386
* [x] The contribution guide has a table on the sections of the docs, and the reference to diataxis framework for their content. Make sure it's updated to reflect the new docs structure!

The new section should look more or less like this:

```text
- Getting started
  - Installing popochiu
  - Get your hands dirty
  - Example resources
  - Tutorials
  - Getting help

- How to develop a game
  - < ... current content of the "Creating a game stub" section ... >

- How-tos and Tutorials
  - <  ... Current content of the  "How to develop a game > Advanced Techniques" section ... >

- The editor handbook
- The engine handbook
- Contribution guide
```

### Exclusions

We are not going to write missing parts now, we are just going to move what we have to make it more intuitive to find information.

Small adjustments may be necessary (linking, phrasing, etc), but keep the scope small so we can release asap.

### Implications

Some link that are outside the docs (old messages, project pages, etc) will be moved. This can break the experience for some users. That's a good moment to investigate if we can create a redirection map so the docs will always appear consistent over time.

stickgrinder:
I have finished this and opened a couple of related issues to further improve the docs.

PR #384, opened 2025-01-16
refs #383: Docs navigation reorganizazion.

stickgrinder:


stickgrinder:
@mapedorr mate, this is good to merge to me.

The scope is making a bit of cleanup in the docs organization and structure, and reflecting the contribution guide dictats.
It's mostly box-moving, no changes to the files if not internal links, page weights and naming.

I added a plugin to enforce redirects (dunno how it works on the production side but I'm pretty sure we'll have some html file with a js redirect in the deployed docs - ok to me, no logic at all!)

The best way to test it is visit the docs locally and see the side-menu. If it makes sense, let's go for it!

stickgrinder:
@mapedorr merging as we agreed on other channels

Issue #234, opened 2024-05-04
Add gizmo handle to Markers


stickgrinder:
### Benefit description

In #232 more solid gizmos has been added to PopochiuClickable objects. Markers are not among them, but for consistency, we should add a gizmo similar to the ones we have for "Walk To Point" and "Look At Point" (#233) to position the markers.

### Solution description

This is as simple as adding a square handle point gizmo to the PopochiuMarker objects.
Of course this gizmo should be configurable in terms of visual aspect, as the clickable ones.

### Exclusions

There is no need to make the Markers gizmo invisible from the toolbar, since it's the sole evident sign that the marker is on the scene.

### Implications

The best option is to have a separate plugin for markers and clickables. At the expense of some duplication, we can manage them without too much branching logic.

stickgrinder:
TODO for this issue:

* [ ] Create a PopochiuMarker class which exposes a `marker_position` or simply `position` **Vector2** property.
* [ ] Create a PopochiuGizmoMarkerPlugin that initializes a gizmo connected to the aforementioned property.
* [ ] Add the configuration for the color and visual properties of the Marker's gizmo
* [ ] Update the factory, and all the functions related to makers to use the new class and property
* [ ] Write a migration to update all the markers in the game rooms
* [ ] Remove legacy marker scenes

stickgrinder:
Leaving some notes to self for the last bits of this implementation:

* The current markers implementation by Anthony is creating a scene file in the game directory. Understand why that's the case (I would say we don't need it for single nodes).
* We can keep a scene (and script?) in case we want to attach a script to markers but I would keep this VERY simple. We may want to extend the base and room object factories to support the creation of simple nodes without folders and scenes. Evaluate from previous point.

stickgrinder:
After talking to @mapedorr we decided to change our approac to solve this.

Relying on Marker2D (leveraging Godot's standards) remains the better options - fight the NIH antipattern!
To give the user a nice and coherent user interface, we'll do this:

- we keep the current implementation by @anthonyirwin82 
- we create marker2d nodes in a hidden state
- the gizmo plugin will show all the markers as gizmos on the scene
- a button in the toolbar will hide and show all the marker gizmos at the same time

Things to solve:

- the gizmos should update the `position` of the Marker2D. The current implementation is not behaving like this. We may want to work this around with a in-memory table, or better, reimplement the plugin on top of the [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern).

[...]

PR #393, opened 2025-02-23
refs #234: markers gizmos

stickgrinder:
This PR is introducing a new popochiu object for markers with its own gizmo.

The goal is to keep the interface consistent.


stickgrinder:
@mapedorr this PR is ready for your review. I forgot to write a migration to make all already present markers invisible (I'll do tomorrow). Also, I mentioned I would have reviewed the code of the toolbar to make it simpler, but I prefer to do it in a maintenance PR, so I won't clutter your review.

Please, take a look at my work and let me know if you see something wrong.
Thanks!

stickgrinder:
@mapedorr migration written. Everything's done it seems.

addons/popochiu/editor/gizmos/gizmo_manager_marker.gd
mapedorr, line 1: If possible, organize this script to use the regions we've defined in the other project scripts: `#region Godot`, `#region Public`, `#region Private`, etc.

mapedorr (approved):
Awesome work!!!

PR #400, opened 2025-04-30
Take into account overlapping areas to check if characters are inside the region

mapedorr:
Also take into account the position and size of the region in order to improve the calculation of the target scaling of the characters inside the region.

Additionally, the definition and access of the`PopochiuCharacter.on_scaling_region` **Dictionary** has been improved.

### DIFF
--- a/addons/popochiu/engine/objects/region/popochiu_region.gd
+++ b/addons/popochiu/engine/objects/region/popochiu_region.gd
@@ -30,6 +30,11 @@ extends Area2D
 ## by [PopochiuRoom] to store the info in its [code].tscn[/code].
 @export var interaction_polygon_position := Vector2.ZERO
 
+var _last_char_pos := Vector2.ZERO
+var _active_characters := {}
+
+@onready var interaction_polygon_node: CollisionPolygon2D = $InteractionPolygon
+
 
 #region Godot ######################################################################################
 func _ready() -> void:
@@ -45,22 +50,22 @@ func _ready() -> void:
 
 		# Ignore assigning the polygon when:
 		if (
-			get_node_or_null("InteractionPolygon") == null # there is no InteractionPolygon node
+			interaction_polygon_node == null # there is no InteractionPolygon node
 			or not get_parent() is Node2D # editing it in the .tscn file of the object directly
 		):
 			return
 
 		# Add interaction polygon to the proper group
-		get_node("InteractionPolygon").add_to_group(
+		interaction_polygon_node.add_to_group(
 			PopochiuEditorHelper.POPOCHIU_OBJECT_POLYGON_GROUP
 		)
 
 		if interaction_polygon.is_empty():
-			interaction_polygon = get_node("InteractionPolygon").polygon
-			interaction_polygon_position = get_node("InteractionPolygon").position
+			interaction_polygon = interaction_polygon_node.polygon
+			interaction_polygon_position = interaction_polygon_node.position
 		else:
-			get_node("InteractionPolygon").polygon = interaction_polygon
-			get_node("InteractionPolygon").position = interaction_polygon_position
+			interaction_polygon_node.polygon = interaction_polygon
+			interaction_polygon_node.position = interaction_polygon_position
 
 		# If we are in the editor, we're done
 		return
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -121,48 +126,71 @@ func _check_scaling(
 	area_rid: RID, area: Area2D, area_shape_index: int, local_shape_index: int, entered: bool
 ):
 	if not (
-		area is PopochiuCharacter 
-		and area.get_node_or_null("ScalingPolygon") 
-		and area_shape_index == area.get_node("ScalingPolygon").get_index()
+		area is PopochiuCharacter
+		and area.get("scaling_polygon")
```
stickgrinder, line 130: Is it worth to have a null-safe check here?

addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -121,48 +126,71 @@ func _check_scaling(
...
+	var character: PopochiuCharacter = area
+	# Track character entry/exit across all shapes
+	if entered:
+		_active_characters[character.script_name] = area
+	else:
+		# Only remove if character is not in any other shape
+		var overlapping := get_overlapping_areas()
+		var still_inside := false
+		for overlapping_area in overlapping:
+			if overlapping_area == character:
```
stickgrinder, line 144: I understand what's going on here but I think the naming of the variables are very hard to get.

Comparing an overlapping area with a character makes little sense. Can't we come up with a different naming? Or better yet, can't this whole loop be implemented using find_key() or get()?

addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -121,48 +126,71 @@ func _check_scaling(
...
-	if scaling:
-		_update_scaling_region(area)
-		area.update_scale()
+		if not still_inside:
+			_active_characters.erase(character.script_name)
+			_clear_scaling_region(character)
+			return
+	
+	if scaling and _active_characters.has(character.script_name):
+		_update_scaling_region(character)
```
stickgrinder, line 154: Same here... maybe this function should be called `_update_character_scaling()` ?

addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -121,48 +126,71 @@ func _check_scaling(
...
+		polygon_y_array.min() + global_position.y + interaction_polygon_node.position.y
+	)
+	var global_bottom: float = (
+		polygon_y_array.max() + global_position.y + interaction_polygon_node.position.y
+	)
+	_last_char_pos = chr.global_position
+
+	var region_height := global_bottom - global_top
+	var position_ratio := (chr.global_position.y - global_top) / region_height
+	position_ratio = clamp(position_ratio, 0.0, 1.0)
```
stickgrinder, line 174: Unsure about where this is coming from and how it is used, because I can't see it in the original diff.
Just check this is not a leftover. Maybe I'll move it down in the code so you focus on calculating the vars that will populate the dictionary and than this one?

addons/popochiu/engine/objects/region/popochiu_region.gd
```diff
@@ -121,48 +126,71 @@ func _check_scaling(
...
+		scale_bottom = self.scale_bottom,
+		scale_max = [self.scale_top, self.scale_bottom].max(),
+		scale_min = [self.scale_top, self.scale_bottom].min(),
+		polygon_top_y = global_top,
+		polygon_bottom_y = global_bottom,
+		target_scale = target_scale
+	}
 
 
 func _clear_scaling_region(chr: PopochiuCharacter) -> void:
```
stickgrinder, line 189: maybe this function can be called "_clear_character_scaling" or "_remove_character_from_region"? I think the former naming is not optimal for readability.

stickgrinder (approved):
I have left some comments that in my opinion may be worth addressing, mostly about the naming of the members of a region, and about a possible leftover.

Feel free to address them or to merge this one as you see fit! No errors or problems detected! ;)

addons/popochiu/engine/objects/region/popochiu_region.gd
mapedorr: You mean for the `area`?

addons/popochiu/engine/objects/region/popochiu_region.gd
mapedorr: Let's see if the new approach feels better.

addons/popochiu/engine/objects/region/popochiu_region.gd
mapedorr: Done.

addons/popochiu/engine/objects/region/popochiu_region.gd
mapedorr: It is used to calculate the target scale for the character. Is part of the fixes needed for the update to Godot 4.4.

addons/popochiu/engine/objects/region/popochiu_region.gd
mapedorr: Done.

addons/popochiu/engine/objects/region/popochiu_region.gd
stickgrinder: Yes, exactly!

Issue #238, opened 2024-05-13
Aseprite Importer: move importers out of the inspector


stickgrinder:
### Benefit description

The Aseprite importers will belong in popups or panels instead of cluttering the inspector.
The current positioning of that interface comes from mimicking PowerQuest placement, which comes from specificities of Unity and - although sensible - is not the best option.

### Solution description

We still need to decide where to place the importers. @stickgrinder will complete this issue as soon as we design the new interaction.

### Exclusions

None

### Implications

To achieve this goal, the Importers will be changed from an Inspector Plugin to an Editor Plugin (like the Gizmos one introduced in #232).

This new architecture, based on a master and many subsidiary plugins, works pretty well to keep things isolated and maintainable. Still, it will require us to decide on an overarching set of standards for how those plugins will interoperate.

I'll leave some comments with options we can see at the moment.

stickgrinder:
To close this issue, the best option is to address #138 first, so we can screenshot the whole new feature. Here is what has to be updated:

* [ ] The Aseprite Importers section, redoing all sshots with the new importer dock
* [ ] The contributions guide, to depict the correct version of the importers (see PR #405 for that)

This is not orthodox, but given the current load of stuff we have to do, it's better to document both issues in a single pass.

PR #401, opened 2025-05-03
refs #238: Aseprite importers now have a dock on their own, out of the inspector.

stickgrinder:


addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock.gd
stickgrinder, line 1: I just changed this function to a public one so I can invoke it deliberately.

Moved at the bottom as for Godot guidelines.

addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock_character.gd
stickgrinder, line 1: Same here.

addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock_room.gd
stickgrinder, line 1: And here.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder, line 1: Removed the placeholder property that I used to substitute with the importer interface, in the inspector.

addons/popochiu/popochiu_plugin.gd
stickgrinder, line 1: Changed the loading from inspector plugin to editor plugin.
I'm reusing the sub-plugin pattern we started with the gizmos here, and I like it: makes Popochiu much more modular!

addons/popochiu/editor/inspector/aseprite_importer_inspector_plugin.gd
stickgrinder, line 1: All the logic in here has been removed, since we now have an EditorPlugin for the importers.

addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock.gd
```diff
@@ -426,4 +397,33 @@ func _show_importer():
...
+#region Public ######################################################################################
+func init():
+	_set_elements_styles()
+	
+	if not PopochiuEditorConfig.aseprite_importer_enabled():
+		_show_info()
+		return
+	
+	# Check access to Aseprite executable
+	var result = _check_aseprite()
```
mapedorr, line 411:
```suggestion
	var result := _check_aseprite()
```

mapedorr (approved):
Everything looks good so far. Please take a look on the comments before merging. The only changes I request, are the ones of the code order. You can skip the others if you want.

Now I'm gonna make some tests, and if I find something I'll let you know.

addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock.gd
stickgrinder: I am so dumb!  :D 

Fixing this!

addons/popochiu/editor/importers/aseprite/docks/aseprite_importer_inspector_dock.gd
```diff
@@ -426,4 +397,33 @@ func _show_importer():
...
+	if result == RESULT_CODE.SUCCESS:
+		_show_importer()
+	else:
+		PopochiuUtils.print_error(RESULT_CODE.get_error_message(result))
+		_show_warning()
+	
+	# Load inspector dock configuration from node
+	var cfg = LOCAL_OBJ_CONFIG.load_config(target_node)
```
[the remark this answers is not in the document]
stickgrinder: Yes, it's a dictionary. It's better to type that variable, and maybe the load_config function too.
Sorry for overlooking this.

addons/popochiu/editor/importers/importers_plugin.gd
```diff
@@ -0,0 +1,138 @@
...
+func _check_for_scene_change() -> void:
+	var root = EditorInterface.get_edited_scene_root()
+	if root:
+		var scene_path = root.scene_file_path
+		# Check if this is a new scene with a path
+		if scene_path != _current_scene_path and not scene_path.is_empty():
+			_current_scene_path = scene_path
+			# Stop the timer and process the scene change
+			_scene_check_timer.stop()
+			_on_scene_changed(root)
```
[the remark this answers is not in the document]
stickgrinder: I like this way more! I overlooked this, but it's the right way to write this function!

addons/popochiu/editor/importers/importers_plugin.gd
```diff
@@ -0,0 +1,138 @@
...
+	var script_path = ""
+	
+	if PopochiuEditorHelper.is_editing_room():
+		script_path = INSPECTOR_DOCK_ROOM
+	elif PopochiuEditorHelper.is_editing_character():
+		script_path = INSPECTOR_DOCK_CHARACTER
+	
+	# Only change the script if needed (avoid unnecessary reloads)
+	if _dock.get_script() != load(script_path):
+		_dock.set_script(load(script_path))
```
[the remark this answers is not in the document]
stickgrinder: I tried to avoid using too much variables for memory usage but it's probably useless optimization and this is actually more optimized! Thanks

mapedorr:
Since you're already modifying these components, I'd like to suggest a few usability improvements (although, in the end, this is just my opinion):

1. For those of us using Godot with a light theme, the buttons have some weird states. You can see them in the screenshot. I know it's annoying, but there is a way to detect which theme is being used and adjust the element colors accordingly. It's not a necessary change for merging, but I want us to keep it in mind for later.
2. By default, **for rooms**, I think the loop option in animations should be disabled. Usually, these animations represent different prop states depending on player interaction. I believe the number of props that loop will always be lower than those that don’t.
3. I think the options section should be collapsed by default. Generally, you only open the options when you want to do something specific (or if something didn’t work as expected). Otherwise, you don’t even look at them.
4. I'd like us to do something to make the Import button stand out more and not blend in so much with the rest. Maybe we could make it a bit "taller", move it to the end, or add an icon. What do you think?

image

stickgrinder:
@mapedorr everything you said is very valuable. I'll comment on each point and propose a course of action:

1. Thanks for noticing, I'll address this for sure! Not an annoyance, I just didn't think about testing with different themes!
2. Makes sense... At the moment we have a configuration flag for that. Do you think we should enforce a default "by type" or that we should expose different flags for each type of importer? I would go with the first one and remove the flag actually... I don't even know if it's actually useful.
3. I'll collapse it and rename it to "Advanced Options" so that people know that you already have sensible defaults.
4. I prefer an icon and maybe a high-contrast look? Since we're no more in the inspector (and we don't have to respect its patterns) I can reuse the patters you used in the main dock.

**Proposal**: what if I create a separated card for those changes? Having this increment on develop will make me able to work on the last importer and I think that use-case will pop up some more UX improvement needs, so I'd like to make a last round of quality-of-like polishing when I have a complete set of features.

Issue #391, opened 2025-02-05
Allow checking if an inventory item was removed.


mapedorr:
### Benefit description

The reason this issue is being created is that **Jezz (#jezztoonz)** (on Discord) mentioned that there is no way for a prop *linked* to an inventory item (`PopochiuProp.link_to_item`) to know that such item was removed from the inventory (e.g., when it is used and no longer makes sense for it to appear in the game) in order to prevent the prop from rendering on screen despite the item not being *inside* the inventory.

Beyond this particular scenario, being able to determine whether an inventory item has been removed could make it easier to program logic that depends on this information. For example, a character could react differently if it is known that a certain inventory item was once there.

### Solution description

Add a `was_removed: bool` property to **PopochiuInventoryItemData** (or **PopochiuInventoryItem**). This way, a prop can check if the inventory item it is linked to _was removed_ from it, and assume it shouldn't display on screen due to that.

Additionally, in **PopochiuProp._ready()** a condition should be added to the `PopochiuUtils.i.is_item_in_inventory(link_to_item)` to include the validation of the new property.

### Exclusions

Can't think of one right now.

### Implications

Can't think of one right now.

stickgrinder:
Adding my two cents on that about the naming of that flag.

What Jezz really wants to know is if the item has ever been collected. Probably the best naming option is `ever_collected`. This would make game scripts much more readable. Compare the following:

```gdscript
if not I.apple.in_inventory and I.apple.was_removed:
```

with:

```gdscript
if not I.apple.in_inventory and I.apple.ever_collected:
```

In the first case, at first glance it seems we're testing the same thing. The second line clarifies the intent a bit more, to me.

PR #402, opened 2025-05-04
refs #391: Adding a flag to know if an item has ever been in the inventory

stickgrinder:
Simple stuff. The value should be stored without adding it explicitely to the data script, since it's a boolean.

Right?


mapedorr (changes requested):
Looks good so far, but in the solution description of the issue I also suggested an additional change.

It should be in the **PopochiuProp._ready()** function. I'm not sure if we want this to always be the case, but the validation should also take into account the `ever_collected` property:

```gdscript
	if link_to_item:
		...
		
		if PopochiuUtils.i.is_item_in_inventory(link_to_item) or ever_collected:
			disable()
```

stickgrinder:
Done, and added a function for coherence to the inventory interface, that I completely missed.

PR #423, opened 2025-08-01
Update popochiu_clickable.gd for touch input fix

Afterworks-Studio:
the value 0 is the output after a one finger touch trigger (also explained in the Popochiu Discord server)


mapedorr (approved):
Thanks!

PR #440, opened 2025-08-23
Fixes #438 - Godot 4.4.1 crashing after plugin installation in Windows

### LINKED ISSUE #418: Crash after scaffolding complex scenes
### Bug description

When creating certain scenes, such as characters or rooms, Godot seems to crash after it finish the scaffolding. Like, everything is there and it works after firing it up again. It just crashes.

###  Steps to reproduce

1. Go to Inspect Main or Room
2. Click on Create room or character
3. Let it finish scaffolding
4. Crash

### Expected behavior

Finish without constantly crashing.

 - **OS type and version**: macOS 15.5 Sequoia
 - **Godot version**: 4.3, 4.4
 - **Plugin version or branch**: 2.0.3

### Additional context

It seems limited to complex scenes because it'll create something like a prop without issue.

### LINKED ISSUE #438: Godot 4.4.1 crashes after creating the first Room
### Bug description

When creating a new project, Godot crashes after the first room is created. After restarting everything looks to work. Could be related to #418 .

###  Steps to reproduce

1. Create a new project, and configure it in the Setup popup.
2. Create a Room.
3. Once the process finishes, Godot will close automatically.

The engine shouldn't close after creating the first room.

 - **OS type and version**: Windows 11.
 - **Godot version**: 4.4.1
 - **Plugin version or branch**: 2.0.3

### Additional context

I created a Godot build with `debug symbols` enabled in order to have more info on what could be happening. Here's the output of the console before the crash:

Image

mapedorr:
The issue was related to deleting the Setup popup after installation. Now, the popup is only hidden, and a reference to it is stored to avoid creating a new one each time developers press the Setup button in Popochiu’s dock.

Additionally, another error was fixed that caused Godot to display a message (or crash on Linux) whenever an object was removed from a Room and from the FileSystem.

This might also fix #418 .

### DIFF
--- a/addons/popochiu/editor/helpers/popochiu_editor_helper.gd
+++ b/addons/popochiu/editor/helpers/popochiu_editor_helper.gd
@@ -137,6 +137,9 @@ static func show_setup(is_welcome := false) -> void:
 		func () -> void:
 			await content.on_close()
 			
+			# The assignment must be done here, since doing it when the ConfirmationDialog is
+			# instantiated causes the engine to crash after trying to create Popochiu objects following
+			# the installation process.
 			_setup_dialog_instance = dialog
 			dialog.hide()
 	)
--- a/addons/popochiu/editor/main_dock/popochiu_row/object_row/room_object_row/popochiu_room_object_row.gd
+++ b/addons/popochiu/editor/main_dock/popochiu_row/object_row/room_object_row/popochiu_room_object_row.gd
@@ -46,8 +46,9 @@ func _remove_from_core(should_save_and_delete := true) -> void:
 			PopochiuResources.Types.WALKABLE_AREA:
 				room_child_to_free = opened_room.get_walkable_area(str(name))
 	
-	# Continue with the deletion flow without saving the scene and deleting the row (that's the
-	# reason why the parent function is called with [false]).
+	# Continue with the deletion flow without saving the scene or deleting the row (that's why the
+	# parent function is called with [false]), since it is removed automatically when the node is
+	# removed from the tree. Saving the scene should happen after that.
 	super(false)
 	
 	# Fix #196: Remove the Node from the Room tree once the folder of the object has been deleted
--- a/addons/popochiu/editor/popups/setup/setup.gd
+++ b/addons/popochiu/editor/popups/setup/setup.gd
@@ -98,7 +98,7 @@ func on_close() -> void:
 	
 	if not PopochiuResources.is_setup_done() or not PopochiuResources.is_gui_set():
 		PopochiuResources.set_data_value("setup", "done", true)
-		await _copy_template(true)
+		await _copy_template()
 
 
 func define_content(show_welcome := false) -> void:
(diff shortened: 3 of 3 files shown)


addons/popochiu/editor/helpers/popochiu_editor_helper.gd
```diff
@@ -122,11 +123,23 @@ static func show_creation_popup(scene: PackedScene, min_size := Vector2i(640, 18
...
 	var content := SETUP_SCENE.instantiate()
 	
 	dialog.title = "Setup"
-	dialog.confirmed.connect(content.on_close)
+	dialog.dialog_hide_on_ok = false
+	dialog.confirmed.connect(
+		func () -> void:
+			await content.on_close()
+			
+			_setup_dialog_instance = dialog
```
stickgrinder, line 143: Why aren't you just initializing the instance in `_setup_dialog_instance` instead of using a local variable that you have to reassign at confirm?

addons/popochiu/editor/main_dock/popochiu_row/object_row/room_object_row/popochiu_room_object_row.gd
```diff
@@ -46,8 +46,9 @@ func _remove_from_core() -> void:
 			PopochiuResources.Types.WALKABLE_AREA:
 				room_child_to_free = opened_room.get_walkable_area(str(name))
 	
-	# Continue with the deletion flow
-	super()
+	# Continue with the deletion flow without saving the scene and deleting the row (that's the
+	# reasin why the parent function is called with [false]).
```
stickgrinder:
```suggestion
	# reason why the parent function is called with [false]).
```

typo

stickgrinder: I would also explain why in case of obj room, we NEVER want the function to delete them and save the scene.

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -285,16 +283,10 @@ func _copy_template(is_first_copy := false) -> void:
...
-	PopochiuGuiTemplatesHelper.copy_gui_template(
+	copy_process_container.show()
+	await PopochiuGuiTemplatesHelper.copy_gui_template(
 		_selected_template.name, _template_copy_progressed, _template_copy_completed
 	)
-	
-	copy_process_container.show()
-	
-	# if true, make the popup visible so devs can see the copy process feedback
-	if is_first_copy:
```
stickgrinder, line 295: If you always want to show the progress popup, I guess the argument can be remove from the function signature, right?

addons/popochiu/editor/helpers/popochiu_editor_helper.gd
mapedorr: Because I'm dumb.

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: You're right. I'll remove the parameter, thanks.

Issue #258, opened 2024-08-13
Improve UX for the Game Setup window


stickgrinder:
### Benefit description

The setup window in 2.0, althought functional, il cluttered and hard to eye-scan.  
Some of the problems we thing we should address:

- The setting of the game should be done pre-selecting a "game type", like "Pixel Art", "High Res", and others if we think it's worth adding more.
- Parameters are pre-set and can be customized only by opening a section in the setup window (like "Advanced options" or "Customize")
- The GUI selection section would scale badly with the addition of more elements. We need to make it sleeker and provide more visual and less textual info.

### Solution description

ATTOW there is no clear solution, just a bunch of ideas that need to be discussed and rendered into a design.

We'll provide a reference design as soon as we have it.

### Exclusions

These changes may highlight the need for resolution-specific game GUIs. They are out of the scope of this issue.

### Implications

We may want to change the default parameters for pixel art games from a **very** retro 320x180 to a retro-inspired 356x200 or slightly larger 384x216 (both 16:9, the second one is very good for large GUIs like the 9verbs, leaving more real estate for the game environment).

We need to decide what a high-res game is. We may also want to provide interfaces in different resolutions, but that will be out of scope (see "Exclusions").

mapedorr:
Looking forward to start creating the HD version of each template!!!

stickgrinder:
This issue is in review.

The best option is to merge this before #374 and rebase that one on top, so we can test the setup with hi-res GUIs before final integration.

The bulk of code is there to keep the UX tidy and solid, as the branching of options and modes makes this pretty articulated.

This will require a good amount of testing and stressing to make sure we don't need more logic to guarantee a good UX.

PR #442, opened 2025-09-21
refs #258: New setup window

stickgrinder:
Too tired to provide a description: see related issue :smile:


addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+
+enum GameResolution {
+	# --- Separator = 0 ---
+	RETRO_NEO_RETRO = 1,
+	# --- Separator = 2 ---
+	RETRO_VGA_4_3 = 3,
+	RETRO_VGA_16_9 = 4,
+	# --- Separator = 5 ---
+	RETRO_CEGA_4_3 = 6,
+	RETRO_CEGA_16_9 = 7,
```
mapedorr, line 32: Clever.

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+var _templates_by_res: Dictionary = {
+	PopochiuGUIInfo.GUITargetRes.LOW_RESOLUTION: [],
+	PopochiuGUIInfo.GUITargetRes.HIGH_RESOLUTION: []
+}
+
+# Template tracking for change detection
+var _current_template_name: String = ""
+var _selected_template_name: String = ""
+var _template_change_confirmed: bool = false
+var _copy_in_progress: bool = false
```
mapedorr, line 78: There's no need to define the variable type if the assignment is done like this in the declaration.

```suggestion
var _current_template_name := ""
var _selected_template_name := ""
var _template_change_confirmed := false
var _copy_in_progress := false
```

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+
+var _is_closing: bool = false
+var _es: EditorSettings = EditorInterface.get_editor_settings()
```
mapedorr, line 81: Same as the previous one.

```suggestion
var _is_closing := false
var _es := EditorInterface.get_editor_settings()
```

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+@onready var lbl_preview_height: Label = %LblPreviewHeight
+# ---- ButtonGroups -------------------------------------------------------------
+@onready var game_type_button_group: ButtonGroup = btn_gametype_retro.button_group
+@onready var gui_button_group: ButtonGroup = btn_gui_type_template.button_group
+# ---- CopyProcess Elements -----------------------------------------------------
+@onready var copy_process_container: PanelContainer = %CopyProcessContainer
 @onready var copy_process_label: Label = %CopyProcessLabel
 @onready var copy_process_bar: ProgressBar = %CopyProcessBar
+# ---- Dialog reference --------------------------------------------------------
+var _dialog_ok_button: Button = null
```
mapedorr: Private properties should go together. Please move this line after 81, or to line 79.

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+	# Style the progress bar
+	copy_process_bar.add_theme_color_override("font_color", get_theme_color("font_color", "Editor"))
+
+
+#endregion
+
+
+#region Signals handlers ######################################################
+
+
```
mapedorr:
```suggestion
```

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+		_update_navigation()
+
+
+func _on_next_pressed() -> void:
+	if wizard_steps.current_tab < wizard_steps.get_tab_count() - 1:
+		wizard_steps.current_tab += 1
+		_update_navigation()
+
+
+func _on_wizard_tab_changed(tab: int) -> void:
```
mapedorr: Add underscore to non used parameters to avoid Godot showing a warning.

```suggestion
func _on_wizard_tab_changed(_tab: int) -> void:
```

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+	# When tab changes, update navigation to trigger validation for the new tab
+	_update_navigation()
+
+
+func _on_resolution_option_changed(index: int) -> void:
```
mapedorr:
```suggestion
func _on_resolution_option_changed(_index: int) -> void:
```

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -4,311 +4,1442 @@ extends Control
...
+	# When any resolution option changes, trigger validation
+	_update_navigation()
+
+
+func _on_custom_field_changed(value = null) -> void:
```
mapedorr:
```suggestion
func _on_custom_field_changed(_value = null) -> void:
```

mapedorr (approved):
You did a terrific job!!!

I left some comments and change suggestions for minor things. I'll test the setup process, but feel free to merge this whenever you want. In case I find something, I'll create the corresponding issue.

addons/popochiu/editor/popups/setup/setup.gd
stickgrinder: Thanks!!! ❤️

addons/popochiu/editor/popups/setup/setup.gd
stickgrinder: Yes, but I'm starting to make it explicit so that it's immediate to read and we don't have mixed declarations. Do you think it's better to always use the shortest format `:=` ?

addons/popochiu/editor/popups/setup/setup.gd
stickgrinder: Same

addons/popochiu/editor/popups/setup/setup.gd
stickgrinder: Yes Sir! 🫡 I'm sorry to be an idiot Sir!

😅

addons/popochiu/editor/popups/setup/setup.gd
stickgrinder: That's what happens when you work in vscode and not in Godot editor.

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: I understand your point. If our goal is to improve readability (or code clarity), then it’s fine to explicitly define the types in the variable declarations.

I’ll try to keep that in mind for the code I write from now on.

Issue #425, opened 2025-08-03
Various improvements to clickables


stickgrinder:
### Benefit description

This is a catchall card related to #417 that sums up simple tasks and easy-wins that affect clickables and/or inventory items.

All that follows can be delivered in a single PR.

### Solution description

* [ ] `enable()`/`disable()` should affect both visibility and interactivity; add `is_enabled()` getter.
* [ ] Add `enable_clickability()`/`disable_clickability()` for clickability only.

PR #443, opened 2025-09-25
refs #425: Added some useful script functions to the clickables.

stickgrinder:
@mapedorr this is a very easy one, maybe you want to review it right away.

### DIFF
--- a/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
+++ b/addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
@@ -243,10 +243,22 @@ func is_enabled() -> bool:
 	return self.visible and self.clickable and self.input_pickable
 
 
+## Make this node clickable.[br][br]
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -217,6 +217,8 @@ func queue_disable() -> Callable:
 ## Hides this Node.
 func disable() -> void:
 	self.visible = false
+	self.clickable = false
+	self.input_pickable = false
```
mapedorr: Using `self.` here is not needed, and since `clickable` has a set method which also changes the value of `input_clickable`, changing that property each time `clickable` changes won't be necessary (this applies also to lines 236, 255, and 267.

stickgrinder: Correct, done.

Issue #428, opened 2025-08-03
Configurable basic animations for the character


stickgrinder:
Child of #417 

Related to: #113 , #129,  #130 and #429

### Benefit description

Currently the engine relies on standard names for idle, walk and talk animations. Being able to set a different name at certain points of the game (or as a whole for a specific character - i.e. `glide` for a ghost or `slide` for a snail?) can be useful. You can support different states like limping walk, sitting/standing positions, etc.

### Solution description

* [ ] Add `idle_animation`, `walk_animation`, `talk_animation` properties in the inspector: String "root names" of the animations to play when idle, walking, or talking (defaults to "idle", "walk", "talk" if unpopulated).
* [ ] The root name should not speciffy directions, they keeps getting added by the engine.

PR #446, opened 2025-09-28
refs #428: Basic animations are now exposed as configurable for each character.

stickgrinder:
This adds:

* Exposed inspector properties, with validation for empty strings, to set the animation names for idle, walk, and talk
* Same property can be set by the scripts, even in queued sequencies
* When the user sets an animation name, in every possible contexts (not just for those three), the system now searches for a snake_case version of the animation before failing, in case the actual string is not found. The reason is aseprite tags are imported and forced to snake case, but the user has no immediate clue. This will make UX better.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -46,6 +46,12 @@ signal obstacle_state_changed(character: PopochiuCharacter)
 
 ## Empty string constant to perform type checks (String is not nullable in GDScript. See #381, #382).
 const EMPTY_STRING = ""
+## Standard idle animation name.
+const STANDARD_IDLE_ANIMATION = "idle"
+## Standard walk animation name.
+const STANDARD_WALK_ANIMATION = "walk"
+## Standard talk animation name.
+const STANDARD_TALK_ANIMATION = "talk"
 
 ## The [Color] in which the dialogue lines of the character are rendered.
 @export var text_color := Color.WHITE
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1001,6 +1029,21 @@ func _translate() -> void:
...
+	_last_requested_animation_label = "null"
+
+	# If the value is empty, return the fallback
+	if value.is_empty():
+		return fallback
+	
+	# Converting to snake_case to match the naming convention in the animation player.
+	# This allows users to input names in various formats, such as camelCase or PascalCase
+	# matching the tags shown in the importer.
+	return value
```
mapedorr, line 1041: Here the comment says it is converting to snake_case, but does that mean it should return `value.to_snake_case()`? Or just the comment is wrong?

stickgrinder: This comment is a leftover. I tried to do it in the setter but it was messing with the actual value (trying to convert it while typing) so I moved this conversion later in execution (gaining a desireable side-effect!)

I'll remove the comment. Thanks

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1001,6 +1029,21 @@ func _translate() -> void:
...
+	# Clear animation cache to force re-evaluation
+	_last_requested_animation_label = "null"
+
+	# If the value is empty, return the fallback
+	if value.is_empty():
+		return fallback
+	
+	# Converting to snake_case to match the naming convention in the animation player.
+	# This allows users to input names in various formats, such as camelCase or PascalCase
+	# matching the tags shown in the importer.
```
stickgrinder:
```suggestion
```

Issue #429, opened 2025-08-03
Temporary override dialog positions for characters


stickgrinder:
### Benefit description

When the GUI makes use of the dialog position point for a character, it may be useful to temporarly override the dialog position or displace it somehow.

Some examples:

* a character "costume" (see #113 ) is taller than the normal character sprite
* the character gets to a walkable are which is near the top of the location and the normal text would be displayed over the sprite, so we want to temporary place the dialog pos below the character's feet
* a character talks while moving or zooming or the camera is doing something that would make the text unreadable, so we want to lock it in position

### Solution description

We'll add the following variables (exposed):

* [ ] `dialog_position_offset`: `Vector2` offset of the dialog box from the character position (added to the normal dialog position).
* [ ] `dialog_position_override`: `Vector2` holding absolute world coordinates for the dialog box position; if set to something other than `Vector2.ZERO` or `null`, it overrides the dialog position entirely.

We'll also add the following methods

* [ ] `reset_dialog_position_override()`: pretty self-explanatory
* [ ] `lock_dialog_position()`: Locks the dialog position so it stays put and visible even if the character moves or is hidden
* [ ] `unlock_dialog_position()`: Unlocks the dialog position so it can be moved or hidden

PR #448, opened 2025-10-01
refs #429: Dialog positions can be altered by scripts.

stickgrinder:


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1005,10 +1077,18 @@ func set_animation_prefix(value: String) -> void:
...
 
-## Getter function. Returns the final destination position from the navigation path, or Vector2(-1, -1) if not moving
+func set_dialog_pos_offset(value: Vector2) -> void:
+	dialog_pos_offset = value
+
+
+func set_dialog_pos_override(value: Vector2) -> void:
+	dialog_pos_override = value
+
+
```
mapedorr: I think these setters could be omitted if they only assign the value without executing additional logic. Is there any specific reason why you'd like to keep them this way?

docs/src/advanced-topics/character-dialog-position-settings.md
```diff
@@ -0,0 +1,86 @@
...
+Lock it in place while moving:
+
+```gdscript
+C.player.walk_to_marker("Parterre")
+C.player.lock_dialog_pos()
+C.player.walk_to_marker("Stage")
+C.player.say("Tze-tze fly!") # displays over the parterre
+C.player.unlock_dialog_pos()
+C.face_down()
+C.player.say("It works.") # displays over the character
```
mapedorr, line 60: I like this one 😄.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: It may be silly but I thought about completeness of the API. Since all setters are public (in that the name don't begin with `_`) and GDScript doesn't really have proper member visibility implementation, I thought about leaving them as part of the API, but it's a bit idiot probably.

You give me the opportunity to clarify that, once I'm done with adding all that stuff to the API I'd like to review all the public interface and decide upon how to expose it coherently (i.e. it's ALWAYS variable assignment? It's always using a set function? It's mixed but with a predictable, intuitive logic?)

In the meantime I can remove these if you prefer.

mapedorr: I use to define setters and getters when one needs to execute additional logic when assigning or reading a value, but I have no issue with leaving them. Maybe the only thing I have to say about using them "always" is that we'll have to make sure all properties through all the scripts have setters and getters for all their properties.

When the time comes to review the entire API, we could discuss what the conditions should be for a property to have setters or getters, or for a function to be public or not.

In the meantime, I'd prefer not to have them (to maintain consistency with the other scripts) 😊.

stickgrinder: Since we don't have taken a final review about the public API yet, I'll follow your advice and the "less code is better code" rule of thumb!

I'm removing these functions. Thanks for your feedback! :)

PR #449, opened 2025-10-12
refs #298: Character Portrait flipping

stickgrinder:
This substitutes and integrates #298 by @edmundito, inheriting the basic logic and:

- exposing _looking_dir by a variadic function "is_facing", which can be useful in game scripts too
- exporting a `flips_when` variable on the GUI component, matching the one in the character, so that the user can configure the flipping logic.

I have strong doubts about the variadic function. It can't be strongly typed and maybe creating two functions:

- `is_facing(dir: Looking)`
- `is_facing_any(dirs: Array[Looking])`

Is a simpler and safer option.
Also, we don't have other variadic functions on the character... Feedback is welcome.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -898,35 +898,34 @@ func get_avatar_for_emotion(emo := EMPTY_STRING) -> Texture:
 	return texture
 
 
-## Checks if the character is currently facing in any of the specified directions.[br][br]
-## You can pass one or more [enum Looking] direction values. This function returns [code]true[/code]
-## if the character's current facing direction ([member _looking_dir]) matches any of the provided
-## directions.[br][br]
+## Checks if the character is currently facing a specific direction.[br][br]
+## Returns [code]true[/code] if the character's current facing direction ([member _looking_dir])
+## matches the provided direction.[br][br]
 ## Example:[br]
 ## [codeblock]
 ## # Check if character is looking left
 ## if character.is_facing(PopochiuCharacter.Looking.LEFT):
 ##     print("Looking left!")
-##
+## [/codeblock]
+func is_facing(dir: Looking) -> bool:
+	return _looking_dir == dir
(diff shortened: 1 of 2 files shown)


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -898,6 +898,37 @@ func get_avatar_for_emotion(emo := EMPTY_STRING) -> Texture:
...
+##
+## # Check if character is looking in any left direction
+## if character.is_facing(
+##     PopochiuCharacter.Looking.LEFT,
+##     PopochiuCharacter.Looking.UP_LEFT,
+##     PopochiuCharacter.Looking.DOWN_LEFT
+## ):
+##     print("Looking leftish!")
+## [/codeblock]
+func is_facing(...directions: Array) -> bool:
```
mapedorr: Hello 4.5 🎉 🎉 🎉

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -898,6 +898,37 @@ func get_avatar_for_emotion(emo := EMPTY_STRING) -> Texture:
...
+	# The character is never facing "no directions"
+	if directions.size() > 0:
+		return false
+	for dir in directions:
+		if not dir is Looking:
```
mapedorr: You could also use this one (which in my opinion has a better reading).

```suggestion
		if dir is not Looking:
```

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: I decided to switch back to two functions: variadic ones are great but:

1. we need to think better where in the gaming API (scripting functions) they have a place and why
2. based on that we may want to update other parts of the engine
3. the current implementation lacks strong typing and I really don't like littering the code with explicit type checks for such little "convenience"

So, welcome 4.5, bye bye 4.5 :smile:

Issue #433, opened 2025-08-03
Extend character following logic and API


stickgrinder:
### Benefit description

Currently the `set_follow_player` variable is available, to make a non-player character follow the player character.  
We can extend a bit the API to allow facing without following, and give convenience methods.

Also there is no way to set _which_ character to start following, that can be really useful in cutscenes.  
Last bit: there is no way to make the player follow another character, which is ideally useful in cutscenes (maybe)?

### Solution description

  * [x] Add `start_facing_character(character)`, `stop_facing_character()` methods that cause the character to face another specified character and continue to turn if they move.
  * [x] `start_following_character(character)`, `stop_following_character()` methods that cause the character to follow another character and continue to follow if they move.
  * [x] Refactor the current `set_follow_player` to use `start_following_character` and `stop_following_character`.
  * [x] Add a migration for those games that are using `follow_player`, changing it by populating the `follow_character` property with the name of the player character.

Consider adding these parameters (coming from #122 - former duplicate that we missed by opening this issue):

* [x] trigger distance (how far the moving character has to walk from its follower before the follower also starts to walk)
* [x] ~stop distance (how close the follower gets to the moving character when it stops walking)~ this is already present, and called `follow_character_offset`... We may need a new naming to tell the starting and stopping offsets.
* [x] ~always face (boolean indicating if the follower should always face the moving character even if this doesn't walk past the trigger distance)~ this is irrelevant: if we want the character to always face another character we can populate both `follow_character` and `face_character`; if we don't want the follower to face the followed if not for moving, we just populate `follow_character`.
* [x] ~face only (face the character but don't walk, just keep the character in sight - useful for non-moving characters)~ this is irrelevant, since we now have a specific and explicit API for facing.
* [x] follow outside the room (boolean telling if the follower has to change room when the player does, or if it's just following it in the single room he's in).

### Implications

The player can ideally be made following another character.
The best option may be to allow setting the following of another character, but for it to be effective only in cutscenes or blocking sequences and ignore it if the player is in control of the player character.

Invoking a move function on a character which is set to follow another one should be possible, but if the leading character moves, the follower will stop his walk and follow the other character.

Following always implies facing.

PR #450, opened 2025-10-13
refs #433: Extend character following API to start and stop following and facing a character

cidneyhamilton:
Add API functions to:

- start following another character
- stop following another character
- start facing another character
- stop facing another character

NPCs who are following a character will always face that character.

### DIFF
--- a/addons/popochiu/migration/migrations/popochiu_migration_9.gd
+++ b/addons/popochiu/migration/migrations/popochiu_migration_9.gd
@@ -1,0 +1,34 @@
+@tool
+class_name PopochiuMigration9
+extends PopochiuMigration
+
+const VERSION = 9
+const DESCRIPTION = "Replace calls to .follow_player to .follow_character(C.player"
+const STEPS = [
+	"Update function name in scripts"
+]
+
+
+#region Virtual ####################################################################################
+## This is code specific for this migration. This should return [code]true[/code] if the migration
+## is successful. This is called from [method do_migration] which checks to make sure the migration
+## should be done before calling this.
+func _do_migration() -> bool:
+	return await PopochiuMigrationHelper.execute_migration_steps(
+		self,
+		[
+			_update_function_name_in_scripts
+		]
+	)
+
+#endregion
+
+
+#region Private ####################################################################################
+
+func _update_function_name_in_scripts() -> Completion:
+	return Completion.DONE if PopochiuMigrationHelper.replace_in_scripts([
+		{from = "follow_player()", to = "follow_character(C.player)"}
+		], ["gui"]) else Completion.IGNORED
+
+#endregion
--- a/addons/popochiu/migration/migrations/popochiu_migration_9.gd.uid
+++ b/addons/popochiu/migration/migrations/popochiu_migration_9.gd.uid
@@ -1,0 +1,1 @@
+uid://epte4dlvu4b6


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
 func get_current_animation() -> String:
 	return _current_animation
 
+func start_facing_character(character: PopochiuCharacter) -> void:
```
stickgrinder, line 1288: Please, add double-hashed comments to all public functions as runtime documentation. Take other public functions around as a reference.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+
+	if self == PopochiuUtils.c.player:
+		# TODO: We might want to have the player follow another character in a cutscene,
+		# but this will have distinct behavior
```
stickgrinder, line 1292: Uhm... I'm unsure we should exclude this altogether... can you explain why do you think we should avoid having the player following another character? Which kind of conflicts do you see and how do you think we should manage those situations?

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+		return
+	
+	if character == null:
+		character = PopochiuUtils.c.player
```
stickgrinder, line 1296: This is a bold assumption, make sure you document it in the docblock.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+		
+	follow_character = character
+	
+	_face_character()
+
+	
+	if not Engine.is_editor_hint():
+		set_process(true)
+
```
stickgrinder, line 1304:
```suggestion


```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+func stop_facing_character() -> void:
+	follow_character = null
+	
+	if not Engine.is_editor_hint():
+		set_process(false)
+
```
stickgrinder:
```suggestion


```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+	
+	if not Engine.is_editor_hint():
+		set_process(true)
+
+func stop_facing_character() -> void:
+	follow_character = null
+	
+	if not Engine.is_editor_hint():
+		set_process(false)
```
stickgrinder, line 1310: I see why you may want to exclude the player character here: this may interfere with a user-triggered walking sequence, which also depends on the processing being active.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
 func get_current_animation() -> String:
 	return _current_animation
 
+func start_facing_character(character: PopochiuCharacter) -> void:
+
```
stickgrinder, line 1289:
```suggestion
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+		set_process(true)
+
+func stop_facing_character() -> void:
+	follow_character = null
+	
+	if not Engine.is_editor_hint():
+		set_process(false)
+	
+func start_following_character(character: PopochiuCharacter) -> void:
+
```
stickgrinder, line 1311:
```suggestion
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1347,20 +1391,23 @@ func _clear_navigation_path() -> void:
...
-		follower_end_position = end_position - follow_player_offset
+		follower_end_position = end_position - follow_character_offset
 	else:
-		follower_end_position = end_position + follow_player_offset
-	walk_to(follower_end_position)
+		follower_end_position = end_position + follow_character_offset
 
+	walk_to(follower_end_position)
 
+# Makes the character face left or right, depending on where the target character is
```
stickgrinder, line 1583:
```suggestion

# Makes the character face left or right, depending on where the target character is
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1111,6 +1106,55 @@ func get_is_visible_in_room() -> bool:
...
+	
+func start_following_character(character: PopochiuCharacter) -> void:
+
+	if self == PopochiuUtils.c.player:
+		# TODO: We might want to have the player follow another character in a cutscene,
+		# but this will have distinct behavior
+		return
+	
+	if character == null:
+		character = PopochiuUtils.c.player
```
stickgrinder, line 1316: Same as above.

addons/popochiu/engine/objects/character/popochiu_character.gd
cidneyhamilton: Thanks! Documentation strings added.

addons/popochiu/engine/objects/character/popochiu_character.gd
cidneyhamilton: Handling the player does add to the scope of the pull request, since we have to handle interaction with normal click-to-move input from the player. I'd rather handle that in a separate issue if someone has a good reason to need this feature.

addons/popochiu/engine/objects/character/popochiu_character.gd
cidneyhamilton: Default behavior mentioned.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: Makes sense, thanks.

Issue #431, opened 2025-08-03
Add alpha value to character and props


stickgrinder:
### Benefit description

Characters and props may be made invisible or traslucent by setting an alpha value, and providing convenient methods for fading effects.

### Solution description

Add this property and methods to props and characters
  * `alpha`: Float (0.0 - 1.0) representing the transparency of the character on the scene.
  * `fade_in(duration: float)` and  `fade_out(duration: float)`: connects to the `alpha` property and make the object hidden or visible after applying a smooth transition.

### Implications

Fading methods should enable and disable the item entirely before and after fading it's transparency.  
Taslucent objects are enable unless explicitely disabled.

PR #457, opened 2025-10-19
refs #431: Alpha fading for characters and props

stickgrinder:


addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -888,6 +893,79 @@ func resume_animation():
...
+## [i]This method is intended to be used inside a [method Popochiu.queue] of instructions.[/i]
+func queue_fade_to(target_alpha: float, duration: float, set_enablement: bool = false) -> Callable:
+	return func(): await fade_to(target_alpha, duration, set_enablement)
+
+
+## Gradually transitions the alpha value from its current value to the specified [param target_alpha]
+## over the specified [param duration] in seconds. The [param target_alpha] value is clamped between
+## [code]0.0[/code] and [code]1.0[/code]. If [param set_enablement] is [code]true[/code], the character
+## will be disabled if the final alpha is 0, or enabled if the final alpha is greater than 0.
+func fade_to(target_alpha: float, duration: float, set_enablement: bool = false) -> void:
```
mapedorr: I would suggest adding a couple of parameters to this function so devs can choose which transition and easing types to apply to the fade tweens.

```suggestion
func fade_to(
	target_alpha: float,
	duration: float,
	set_enablement: bool = false,
	trans := Tween.TransitionType.TRANS_LINEAR,
	ease := Tween.EaseType.EASE_IN_OUT
) -> void:
```

mapedorr (approved):
Just left a suggestion. Feel free to apply it if you think it makes sense, otherwise you can merge the branch. Of course you'll need to apply it to the methods in the **PopochiuProp** class.

Additionally, I'm not sure if the fading functions should go in the **PopochiuClickable** class since both characters and props will make use of them. I guess you decided to separate them since it makes no sense to use them on hotspots, but let me know what you think about.

addons/popochiu/engine/objects/character/popochiu_character.gd
stickgrinder: Added, with logic to both classes.

About the code duplication, I'm not a fan but inheritance is pretty terrible in these cases; and this is not an isolated case, sadly.

Composition via traits/mixins would be an incredible addition to GDScript. It seems they consider it valuable but not a priority for now.

One thing we can do is to review all the cases we have for common logic and decide on the proper architecture once we're done: we may have a common ancestor between Clickable and "visible" clickables for example.
But we'd also need something in between clickables and inventory items which share some API.

All in all I think we should go in two steps: we'll accept some duplicated code while making the API complete and - at a second stage - we will review the whole API and decide the inheritance chain.
Maybe we can understand if decorators are a viable solution, even if I feel like they can be just "moving shit around".

Or we may decide to wait for traits to be available and refactor this code as god intended. That would be a real improvement and a sensible rewriting.

PR #461, opened 2025-11-16
refs #433: New character following functions

stickgrinder:


addons/popochiu/engine/objects/room/popochiu_room_data.gd
```diff
@@ -120,13 +120,24 @@ func save_children_states() -> void:
 						continue
 					# ---------------------------------------------------------------- Fix #320 ----
 					
-					var script_path := scene_path.replace("tscn", "gd")
-					if not FileAccess.file_exists(script_path):
+					var packed_scene: PackedScene = load(scene_path)
+					if not packed_scene or not packed_scene is PackedScene:
```
mapedorr: I don't think this validation, along with lines 125 and 126, are necessary.

addons/popochiu/engine/objects/room/popochiu_room_data.gd
```diff
@@ -120,13 +120,24 @@ func save_children_states() -> void:
...
 						folder_name = dir.get_next()
 						continue
 					
-					var node: Node2D = load(script_path).new()
-					node.script_name = folder_name.to_pascal_case()
+					# Instantiate the full scene so exported data (like walkable polygons) is already baked.
+					# Loading only the script left interaction_polygon empty, so the first time we entered
+					# a room the stored state overwrote its walkable areas with blank data. Grabbing the
+					# packed scene preserves those serialized outlines until the room can save itself.
+					var node: Node = packed_scene.instantiate()
```
mapedorr, line 138: I get the point of doing this, but there’s something I keep thinking about: as far as I remember, the baked info for navigation and interaction polygons is saved in the Room, not in each WalkableArea or each Prop/Hotspot. If this fixes the bug, we can leave it as is, but maybe we should check whether we’re actually handling that info correctly. Maybe the more logical solution would be to save that data in the corresponding WalkableArea or Clickable scene when the Room gets modified.

stickgrinder: It is working as-is, but I think this comes from the fact that when the room gets intialized, the walkable area is rebaked based on its geometry + all the obstacles.

This way, saving the geometry of the WA seems to be enough to get it back in a working state.
I admit I'm not 100% sure about the lifecycle at the base of this fix.

Probably it requires more tests.

addons/popochiu/engine/objects/room/popochiu_room_data.gd
stickgrinder: You're right...

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -243,6 +251,13 @@ func room_readied(room: PopochiuRoom) -> void:
 		
 		if not chr: continue
 		
+		# If character is already in another room's tree, remove it first
+		# This can happen when the same character instance is saved in multiple rooms
+		if chr.is_inside_tree():
+			var parent = chr.get_parent()
```
mapedorr:
```suggestion
			var parent := chr.get_parent()
```

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -256,9 +271,43 @@ func room_readied(room: PopochiuRoom) -> void:
...
+		# This must happen after add_character() so the character is in the tree
+		if chr_dic.has("follow_character") and chr_dic.follow_character:
+			chr.follow_character = chr_dic.follow_character
+
 		current.add_character(chr)
-	
+
+		# Restore face link after character is added to the room
+		if not chr.face_character.is_empty():
+			var faced_chr = PopochiuUtils.c.get_character(chr.face_character)
```
mapedorr:
```suggestion
			var faced_chr := PopochiuUtils.c.get_character(chr.face_character)
```

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -256,9 +271,43 @@ func room_readied(room: PopochiuRoom) -> void:
...
+			if is_instance_valid(faced_chr) and faced_chr.is_inside_tree():
+				chr.start_facing_character(faced_chr)
+
+		# Restore follow link after character is added to the room
+		if not chr.follow_character.is_empty():
+			var followed_chr = PopochiuUtils.c.get_character(chr.follow_character)
```
mapedorr:
```suggestion
			var followed_chr := PopochiuUtils.c.get_character(chr.follow_character)
```

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -352,3 +405,136 @@ func set_current(value: PopochiuRoom) -> void:
...
+		_collect_followers_recursive(follower.script_name, follower_map, processed)
+
+
+# Adds characters that followed across rooms to the current room.
+# Positions them relative to their followed character and re-establishes the follow link.
+func _add_cross_room_followers() -> void:
+	if _pending_cross_room_followers.is_empty():
+		return
+	
+	for entry in _pending_cross_room_followers:
```
mapedorr:
```suggestion
	for entry: Dictionary in _pending_cross_room_followers:
```

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -352,3 +405,136 @@ func set_current(value: PopochiuRoom) -> void:
...
+		# Stop any ongoing movement before repositioning
+		follower.stop_walking()
+		
+		# Position the follower relative to the followed character AFTER add_character(),
+		# because add_character() calls update_position() which would overwrite with the
+		# stale _buffered_position from the previous room.
+		var target_position: Vector2 = followed.position + offset
+		follower.position = target_position
+		# Sync the buffered position so it doesn't jump back when movement starts
+		follower._sync_buffered_position()
```
mapedorr: Following our guidelines, we shouldn't call private/virtual methods from outside objects directly. I don't want to be a bother, but in this case there should be a public method `sync_buffered_position()` so this call can be made.

addons/popochiu/engine/objects/room/popochiu_room.gd
```diff
@@ -212,8 +212,21 @@ func add_character(chr: PopochiuCharacter) -> void:
 	chr.idle()
 
 
-## Removes [param chr] the [b]$Characters[/b] node without destroying it.
+## Removes [param chr] from the [b]$Characters[/b] node without destroying it.
+## [br][br]
+## [b]Note:[/b] This removal persists across room transitions. A removed character will not be restored
+## when returning to the room. If you want the character to reappear on subsequent visits, either
+## use [method add_character] in the room's [method PopochiuRoom._on_room_entered] callback
+## or hide the character instead ([code]character.visible = false[/code]).
```
mapedorr: I think is better to suggest calling `disable()` instead of just hiding the character because it won't prevent collisions (for interactions) from working.

```suggestion
## or hide the character instead ([code]character.disable()[/code]).
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -64,14 +69,37 @@ const STANDARD_TALK_ANIMATION = "talk"
...
+## Set to [code]Vector2.ZERO[/code] to make the follower move on every step (no threshold).
+@export var follow_character_threshold := Vector2(35, 10)
+## When [code]true[/code], this character will be automatically transferred to the target room when
+## the followed character changes rooms. The follower will appear at the followed character's
+## position plus [member follow_character_offset]. Chain-following is supported: if A follows B
+## and B follows C, and C changes room, both A and B will transfer.
+@export var follow_character_outside_room := false
+## The [member PopochiuCharacter.script_name] of the character that this character should
+## continuously face. Set this in the inspector to have the character automatically face another
+## character at runtime.
```
mapedorr: The description that should go here is the one of the `avatars` property.

```suggestion
## Array of [Dictionary] where each element has [code]{ emotion: String, avatar: Texture }[/code].
## You can use this to define which [Texture] to use as avatar for the character when it speaks
## using a specific emotion.
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -64,14 +69,37 @@ const STANDARD_TALK_ANIMATION = "talk"
...
-## The offset between the character being followed and this character when it follows the former one
-@export var follow_character_offset := Vector2(20, 0)
+## The [member PopochiuCharacter.script_name] of the character that this character should follow
+## when it moves through the room. Set this in the inspector to have the character automatically
+## follow another character at runtime.
+@export var face_character := ""
 ## Array of [Dictionary] where each element has [code]{ emotion: String, avatar: Texture }[/code].
 ## You can use this to define which [Texture] to use as avatar for the character when it speaks
 ## using a specific emotion.
+@export var follow_character := ""
```
mapedorr, line 79: The description of this property was misplaced.

```suggestion
## The [member PopochiuCharacter.script_name] of the character that this character should
## continuously face. Set this in the inspector to have the character automatically face another
## character at runtime.
@export var follow_character := ""
```

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -250,17 +290,32 @@ func _ready():
...
+	if _current_faced_character:
+		_face_character(_current_faced_character)
+
+
+func _exit_tree() -> void:
+	# Safety cleanup: disconnect all follow/face signals to prevent memory leaks.
+	if _current_followed_character:
+		stop_following_character()
+	if _current_faced_character:
+		stop_facing_character()
```
mapedorr, line 316: image

addons/popochiu/engine/interfaces/i_room.gd
stickgrinder: You're not a bother, I totally forgot to review this!

Thanks!!!

addons/popochiu/engine/objects/room/popochiu_room_data.gd
stickgrinder: Before merging this, I'd like to clarify with you one thing about your comment.

Are you suggesting we stop saving interaction/navigations polygons in room properties? This was necessary for serialization (or we would be unable to recreate the walkable areas when we load the room).

Are you suggesting we move these serializable vars into their respective objects, changing the current architecture?

mapedorr: Yes, but to be honest, I'm not sure we're storing that info in the `.tscn` files of each object. The other day, while testing something, I noticed that opening the `.tscn` file of a **WalkableArea** showed the same polygon set as the room's scene.

We can keep this as it is for now and, if necessary, have a discussion later about where this information should be stored.

Issue #374, opened 2025-01-01
Provide current GUIs for hi-res games


stickgrinder:
### Dependencies

This issue **follows** #373
This issue **precedes** #258 

### Benefit description

Currently, Popochiu doesn't provide usable GUIs for high-resolution games. The experimental scaling is sub-optimal for those who want a high-res visual rendition, and it requires some trickery at runtime.

Those who want to create a high-resolution game should be able to have at least a valid version of the SimpleClick GUI at their service (see #373).

Despite the 9-Verbs (Lucas) and the Icon-Bar (Sierra) interfaces are probably bound to be used in retro-inspired games only, should providing a high-resolution friendly version of both is feasible with a reasonable effort, they have to be considered in the scope of this issue.

### Solution description

* [ ] A high-resolution theme for the GUIs must be provided, as neutral as the ones we have already available
* [ ] High-resolution icons and assets have to be created. We can take inspiration from #373 or from the current icons, or again design a set of new assets that play well with Popochiu
* [ ] We must select one or more fonts to distribute with high-res GUI templates.
* [ ] These GUIs have to be made available for selection in the setup window (see #258, which is related to this).

### Exclusions

Since #258 will be included in 2.1 alongside the new high-res GUIs, this issue will NOT include big changes to the setup window. We can still take inspiration from the "custom" step detailed in #258 and rearrange the icon to select the GUI on two rows in the current setup window, removing the wall-of-text for description.

The best option is any way to keep this issue cleanly separated from #258.

PR #473, opened 2025-12-24
refs #374: High-res GUIs

mapedorr:
Finally Popochiu will have High-res GUIs for each of the templates it offers: Simple Click, 9 Verbs and Sierra (Action Bar)!!!

This PR enables 3 new GUI templates for games created with Popochiu. Each template makes use of (temporary) SVG images for icons and cursors. All of them share the same style for buttons and panels.

Additionally, some issues were fixed in the **Setup popup**:

* The style of buttons and other popup elements was being overwritten in the scene (_.tscn_ file) every time it was opened, because the methods that define those styles were being called in the `_ready()` function. Since this is a `@tool` script, those changes were being saved to disk. Now, buttons and other elements keep Godot's default styles, while only some properties are overridden when the popup is opened after pressing the Setup button.
* The game window resolution was not being selected correctly in the step 2 dropdown.
* Changes made in the popup were being saved to Settings even when closing the popup (this behavior existed before), but this caused issues when switching templates, since the copy process was happening without showing the progress bar, which could lead to usability problems.
* When no game type was selected (retro or modern), the game resolution could be saved as `(0, 0)`, which caused issues when running the game or opening scenes. A default resolution is now set depending on the game type.
* The tooltip and button states in the template selection step were not behaving clearly when the template change was canceled. Button states and styles, along with the info shown in the tooltip, have now been improved.

### DIFF
--- a/addons/popochiu/editor/popups/setup/setup.gd
+++ b/addons/popochiu/editor/popups/setup/setup.gd
@@ -762,7 +762,7 @@ func _set_wizard_game_resolution() -> void:
 
 	match _game_type:
 		GameType.RETRO:
-			match opt_res_retro.selected:
+			match opt_res_retro.get_item_id(opt_res_retro.selected):
 				GameResolution.RETRO_NEO_RETRO:
 					_game_resolution = Vector2(384, 216)
 				GameResolution.RETRO_VGA_4_3:
--- a/addons/popochiu/plugin.cfg
+++ b/addons/popochiu/plugin.cfg
@@ -4,5 +4,5 @@ name="Popochiu"
 description="[en] Point n' click games engine for Godot.
 [es] Motor para crear juegos de aventura gráfica en Godot."
 author="Carenalgas Dev Team"
-version="2.0.3"
+version="2.1"
 script="popochiu_plugin.gd"
--- a/addons/popochiu/editor/popups/setup/setup.tscn
+++ b/addons/popochiu/editor/popups/setup/setup.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=12 format=3 uid="uid://5g3dk8ky51g5"]
+[gd_scene format=3 uid="uid://5g3dk8ky51g5"]
 
 [ext_resource type="Script" uid="uid://ccsu88s0qt43b" path="res://addons/popochiu/editor/popups/setup/setup.gd" id="1_htti4"]
 [ext_resource type="Texture2D" uid="uid://b31n8y436g5bg" path="res://addons/popochiu/icons/logo-dark.svg" id="2_ba3yh"]
@@ -25,7 +25,7 @@ resource_name = "GameGUIBtnGroup"
 
 [sub_resource type="StyleBoxLine" id="StyleBoxLine_htti4"]
 
-[node name="Setup" type="Control"]
+[node name="Setup" type="Control" unique_id=1129574411]
 layout_mode = 3
 anchors_preset = 15
 anchor_right = 1.0
@@ -33,8 +33,9 @@ anchor_bottom = 1.0
 grow_horizontal = 2
 grow_vertical = 2
 script = ExtResource("1_htti4")
+metadata/_edit_vertical_guides_ = [-132.0]
 
-[node name="WizardContainer" type="PanelContainer" parent="."]
+[node name="WizardContainer" type="PanelContainer" parent="." unique_id=1014874095]
 unique_name_in_owner = true
 custom_minimum_size = Vector2(520, 600)
 layout_mode = 1
@@ -44,63 +45,64 @@ anchor_bottom = 1.0
 grow_horizontal = 2
 grow_vertical = 2
 
-[node name="Rows" type="VBoxContainer" parent="WizardContainer"]
+[node name="Rows" type="VBoxContainer" parent="WizardContainer" unique_id=547866189]
 layout_mode = 2
 theme_override_constants/separation = 32
 
-[node name="WizardSteps" type="TabContainer" parent="WizardContainer/Rows"]
+[node name="WizardSteps" type="TabContainer" parent="WizardContainer/Rows" unique_id=122850428]
 unique_name_in_owner = true
 layout_mode = 2
 theme_override_styles/panel = SubResource("StyleBoxEmpty_4duov")
-current_tab = 0
+current_tab = 2
 tabs_visible = false
 
-[node name="StepType" type="PanelContainer" parent="WizardContainer/Rows/WizardSteps"]
+[node name="StepType" type="PanelContainer" parent="WizardContainer/Rows/WizardSteps" unique_id=548601569]
+visible = false
 layout_mode = 2
 theme_override_styles/panel = SubResource("StyleBoxEmpty_vh03r")
 metadata/_tab_index = 0
 
-[node name="Rows" type="VBoxContainer" parent="WizardContainer/Rows/WizardSteps/StepType"]
+[node name="Rows" type="VBoxContainer" parent="WizardContainer/Rows/WizardSteps/StepType" unique_id=56918160]
 layout_mode = 2
 theme_override_constants/separation = 16
 
-[node name="SpacerTop" type="Panel" parent="WizardContainer/Rows/WizardSteps/StepType/Rows"]
+[node name="SpacerTop" type="Panel" parent="WizardContainer/Rows/WizardSteps/StepType/Rows" unique_id=946070364]
 custom_minimum_size = Vector2(0, 1)
 layout_mode = 2
 theme_override_styles/panel = SubResource("StyleBoxEmpty_glykx")
 
-[node name="ContainerLogo" type="CenterContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows"]
+[node name="ContainerLogo" type="CenterContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows" unique_id=1682825474]
 layout_mode = 2
 
-[node name="TextureRect" type="TextureRect" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ContainerLogo"]
+[node name="TextureRect" type="TextureRect" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ContainerLogo" unique_id=1518138368]
 custom_minimum_size = Vector2(160, 0)
 layout_mode = 2
 texture = ExtResource("2_ba3yh")
 expand_mode = 5
 
-[node name="Spacer" type="HSeparator" parent="WizardContainer/Rows/WizardSteps/StepType/Rows"]
+[node name="Spacer" type="HSeparator" parent="WizardContainer/Rows/WizardSteps/StepType/Rows" unique_id=2107316917]
 custom_minimum_size = Vector2(0, 1)
 layout_mode = 2
 theme_override_styles/separator = SubResource("StyleBoxEmpty_glykx")
 
-[node name="LblCTAType" type="Label" parent="WizardContainer/Rows/WizardSteps/StepType/Rows"]
+[node name="LblCTAType" type="Label" parent="WizardContainer/Rows/WizardSteps/StepType/Rows" unique_id=1507825925]
 unique_name_in_owner = true
 layout_mode = 2
 text = "What kind of game do you want to make?"
 horizontal_alignment = 1
 justification_flags = 160
 
-[node name="ButtonGroup" type="MarginContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows"]
+[node name="ButtonGroup" type="MarginContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows" unique_id=342693757]
 custom_minimum_size = Vector2(0, 96)
 layout_mode = 2
 theme_override_constants/margin_left = 80
 theme_override_constants/margin_right = 80
 
-[node name="Columns" type="HBoxContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ButtonGroup"]
+[node name="Columns" type="HBoxContainer" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ButtonGroup" unique_id=1407004376]
 layout_mode = 2
 theme_override_constants/separation = 48
 
-[node name="BtnTypeRetro" type="Button" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ButtonGroup/Columns"]
+[node name="BtnTypeRetro" type="Button" parent="WizardContainer/Rows/WizardSteps/StepType/Rows/ButtonGroup/Columns" unique_id=687792552]
 unique_name_in_owner = true
 layout_mode = 2
 size_flags_horizontal = 3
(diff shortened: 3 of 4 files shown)


addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -85,31 +85,31 @@ var _es: EditorSettings = EditorInterface.get_editor_settings()
...
 @onready var lbl_cta_res: Label = %LblCTARes
 @onready var lbl_cta_preview_scale: Label = %LblCTAPreviewScale
 @onready var lbl_cta_gui: Label = %LblCTAUI
-# ---- Game Type Step ----------------------------------------------------------
-@onready var btn_gametype_retro: Button = %BtnTypeRetro
-@onready var btn_gametype_modern: Button = %BtnTypeModern
-# ---- Resolution Step ---------------------------------------------------------
+# ---- Game Type Step ------------------------------------------------------------------------------
+@onready var btn_type_retro: Button = %BtnTypeRetro
+@onready var btn_type_modern: Button = %BtnTypeModern
```
stickgrinder, line 111: May I ask why?

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -240,6 +247,18 @@ func on_about_to_popup() -> void:
...
+	# doing it in the _ready() function results in control styles being modified every time the scene
+	# is opened in the Editor. Until we find an elegant way to determine if the scene is being edited
+	# or is an instance created by pressing the Setup button, styling will be applied before showing
+	# the popup.
+	_style_navigation_buttons()
+	_style_tooltips()
+	_style_selection_buttons()
+	_style_separators()
+	_style_underfield_labels()
+	_style_progress_container()
```
stickgrinder, line 261: Great! I suck at that!

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -342,9 +333,47 @@ func define_content() -> void:
 	_select_current_template()
 
 
+func on_confirm() -> void:
```
stickgrinder, line 336: Can you explain why did you move that to on_confirm()? It sounds perfectly legit, I want to learn.

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -719,6 +773,8 @@ func _set_wizard_game_resolution() -> void:
 					_game_resolution = Vector2(240, 180)
 				GameResolution.RETRO_CEGA_16_9:
 					_game_resolution = Vector2(320, 180)
+				_:
+					_game_resolution = Vector2(356, 200)
```
stickgrinder, line 777: We should never be in this situation, but ok as a safety net, I guess...

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -1179,73 +1241,50 @@ func _style_separators() -> void:
...
-	btn_focus_style.bg_color = btn_bg_color.darkened(0.5)
-	btn_focus_style.border_color = get_theme_color("selection_color", "Editor")
-	btn_focus_style.set_border_width_all(2)
-
-	for button in [btn_gametype_retro, btn_gametype_modern, btn_gui_type_template]:
+	# Determine if the editor is using a Light or Dark color
+	var editor_base_color: Color = EditorInterface.get_editor_settings().get_setting(
+		"interface/theme/base_color"
+	)
+	var is_light := editor_base_color.get_luminance() > 0.5
```
stickgrinder, line 1278: Is this really the only way to address this problem? 🤦 

I mean, isn't there any Editor Theme variable we can use that is automatically set from the editor in a way that makes this coherent?

addons/popochiu/editor/popups/setup/setup.gd
```diff
@@ -1254,40 +1293,68 @@ func _style_selection_buttons() -> void:
...
-		button.add_theme_color_override("font_color", accent_color)
-		button.add_theme_color_override("font_hover_color", accent_color.lightened(0.3))
-		button.add_theme_color_override("font_pressed_color", accent_color.lightened(0.6))
-
 
 func _style_progress_container() -> void:
 	# Apply consistent theming to progress container elements
 	var base_font_size: int = get_theme_font_size("main_size", "EditorFonts")
-
+
```
stickgrinder:
```suggestion

```

stickgrinder: Removed whitespace

addons/popochiu/engine/objects/gui/templates/9_verb_high_res/9_verb_high_res_gui_info.tres
```diff
@@ -0,0 +1,16 @@
+[gd_resource type="Resource" script_class="PopochiuGUIInfo" load_steps=3 format=3 uid="uid://cdl3s004512dx"]
+
+[ext_resource type="Texture2D" uid="uid://b8npdr2hnbf1h" path="res://addons/popochiu/icons/btn_setup_gui_9verbs.svg" id="1_jx7rp"]
+[ext_resource type="Script" uid="uid://dx8nbjp33mba4" path="res://addons/popochiu/engine/others/popochiu_gui_info.gd" id="2_subcu"]
+
+[resource]
+script = ExtResource("2_subcu")
+title = "9 Verb (High-res)"
```
stickgrinder: Since we selected the "modern" game style, does it actually matters that we specify "High-res" in the name of the GUI? It may be more "magical" to select the game type, than the GUI style and see it adapted and properly placed, no?

addons/popochiu/engine/objects/gui/templates/sierra_high_res/images/btn_interact.svg
stickgrinder, line 1: I have a small suggestion for this hand icon (I know it would reverberate on other icons like the actionbar).

I'll send my variant on discord.

addons/popochiu/engine/objects/gui/templates/sierra_high_res/images/btn_walk.svg
stickgrinder, line 1: I like this one SO much!

addons/popochiu/engine/objects/gui/templates/sierra_high_res/sierra_high_res_gui_info.tres
```diff
@@ -0,0 +1,13 @@
+[gd_resource type="Resource" script_class="PopochiuGUIInfo" load_steps=3 format=3 uid="uid://d2wkjty56oqky"]
+
+[ext_resource type="Texture2D" uid="uid://bt3bqfem8qhle" path="res://addons/popochiu/icons/btn_setup_gui_actionbar.svg" id="1_at2o4"]
+[ext_resource type="Script" uid="uid://dx8nbjp33mba4" path="res://addons/popochiu/engine/others/popochiu_gui_info.gd" id="2_ra6a5"]
+
+[resource]
+script = ExtResource("2_ra6a5")
+title = "Action Bar (High-res)"
```
stickgrinder: Same as above, I would remove the specification. It sounds like the "proper" ones are the low-res.

addons/popochiu/engine/objects/gui/templates/simple_click_high_res/simple_click_high_res_gui_info.tres
```diff
@@ -0,0 +1,16 @@
+[gd_resource type="Resource" script_class="PopochiuGUIInfo" load_steps=3 format=3 uid="uid://ub60t3miblvg"]
+
+[ext_resource type="Texture2D" uid="uid://4kgp2a65o7a0" path="res://addons/popochiu/icons/btn_setup_gui_simpleclick.svg" id="1_cayax"]
+[ext_resource type="Script" uid="uid://dx8nbjp33mba4" path="res://addons/popochiu/engine/others/popochiu_gui_info.gd" id="2_jg2de"]
+
+[resource]
+script = ExtResource("2_jg2de")
+title = "Simple Click (High-res)"
```
stickgrinder: Ant that's the last 😄 I promise

addons/popochiu/engine/templates/gui/9_verb_high_res_commands_template.gd
stickgrinder, line 1: I think we talked about possible renundant parts like this, and decided that this would address divergence, right?

addons/popochiu/migration/migrations/popochiu_migration_4.gd
```diff
@@ -110,11 +110,7 @@ func _add_new_components() -> Completion:
 
 func _update_save_and_load_popup() -> Completion:
 	# Update the save_and_load_popup.gd in the game's gui folder
-	DirAccess.rename_absolute(
-		ADDON_SAVE_AND_LOAD_POPUP_SCRIPT,
-		GAME_SAVE_AND_LOAD_POPUP_SCRIPT
-	)
-	
+	DirAccess.copy_absolute(ADDON_SAVE_AND_LOAD_POPUP_SCRIPT, GAME_SAVE_AND_LOAD_POPUP_SCRIPT)
```
stickgrinder, line 115: What happened there? 😶

addons/popochiu/migration/migrations/popochiu_migration_6.gd
```diff
@@ -42,7 +42,9 @@ func _add_animation_players_to_inventory_items() -> Completion:
 	
 	while file_name != "":
 		if dir.current_is_dir():
-			var item_scene_path: String = PopochiuResources.get_data_value("inventory_items", file_name.to_snake_case(), null)
+			var item_scene_path: String = PopochiuResources.get_data_value(
```
stickgrinder, line 47: Is this addressing #474 by any chance? If so I would put a comment over it.

addons/popochiu/migration/migrations/popochiu_migration_8.gd
```diff
@@ -384,24 +433,32 @@ func _get_character_scripts(characters_path: String) -> Array[String]:
...
+					"Migration %d: Character script not found: %s" % [
+						VERSION, character_script_path
+					]
+				)
 		
 		character_name = dir.get_next()
 	
 	return character_scripts
-    
+
```
stickgrinder:
```suggestion

```

whitespace

addons/popochiu/plugin.cfg
```diff
@@ -4,5 +4,5 @@ name="Popochiu"
 description="[en] Point n' click games engine for Godot.
 [es] Motor para crear juegos de aventura gráfica en Godot."
 author="Carenalgas Dev Team"
-version="2.0.3"
+version="2.1"
```
stickgrinder: We generally have this into a dedicated release commit, but let's put it here after all this work ;)

stickgrinder (commented):
Most of the comments are questions, just a few suggestions.

Great work! Can't wait to merge it 👍

addons/popochiu/engine/objects/gui/templates/sierra_high_res/images/btn_interact.svg
stickgrinder: Image

This is a **very** quick and dirty compaison for you to evaluate.

In small factors the inward curve below the thumb makes it rensamble a duck's foot to my eye.
Of course, that's not vital, just a suggestion.

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: I think is a good practice that the name of the variable that points to a Node in the tree matches the name of such node (but in snake_case). It makes easier to "find" those nodes on the scene when needed.

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: Having all the logic in `on_close()` was causing some tasks to run when the popup was closed (using the X button) without being confirmed. Certain changes, like switching the GUI style, should only happen when the popup is explicitly confirmed, to avoid running copy logic outside of the expected flow for developers (which includes the progress bar).

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: Actually it was happening a lot to me. If you opened the Setup popup and then closed it with the X button, the resolution of the game was changed to `(0, 0)` in some cases. I can't remember if it was when you changed the game type and then moved to the game resolution step.

addons/popochiu/editor/popups/setup/setup.gd
mapedorr: Well... now that you mention it, we could use the `interface/theme/preset` property, which has the name of the current theme. Even though using that name might not be accurate (for example, I know it's an exaggerated scenario, but what if someone has the Light theme enabled and customized it to use dark colors?).

I asked ChatGPT and its suggestion was similar to this one (using the luminance, or asking for specific colors).

addons/popochiu/engine/objects/gui/templates/9_verb_high_res/9_verb_high_res_gui_info.tres
mapedorr: Agree. I'll change that name.

addons/popochiu/engine/objects/gui/templates/sierra_high_res/sierra_high_res_gui_info.tres
mapedorr: I'll change it.

addons/popochiu/engine/objects/gui/templates/simple_click_high_res/simple_click_high_res_gui_info.tres
mapedorr: This one will be renamed too!

addons/popochiu/engine/objects/gui/templates/sierra_high_res/images/btn_interact.svg
mapedorr: Ok. I can change it. In any case, these icons are temporary. We should keep these considerations in mind when we're working on our own icons.

addons/popochiu/engine/templates/gui/9_verb_high_res_commands_template.gd
mapedorr: If I'm not mistaken, what you're concerned about is whether this script is a copy of the one used by the low-res interface, right?

If that's the case, I agree they should be the same. I can spend some time reviewing the GUI copy logic so that, as a fallback, it looks for the low-res version of the theme template (or a fallback template file that could be used by any GUI).

If that's not the case, let me know so we can discuss about it.

addons/popochiu/migration/migrations/popochiu_migration_6.gd
mapedorr: Yes, it is. By the time I found this issue, the issue hadn't been created yet, and it didn't occur to me (because stupidity) to create one to show that it had already been fixed in my branch.

I'll leave a comment pointing that out.

addons/popochiu/migration/migrations/popochiu_migration_4.gd
mapedorr: Don't know. I was surprised too. I don't think this is an error we've been carrying over since the last release (otherwise the plugin wouldn't work at all). I'm not sure if it was an internal change in Godot between versions 4.3 and 4.5, but this fixes it.

addons/popochiu/plugin.cfg
mapedorr: I did this because I was having issues when testing things with various editors opened. But we should follow our workflow (tradition) and make the version update in a dedicated release commit. I'm gonna keep the 2.0.3 in my branch.

PR #483, opened 2026-02-16
Fixes #481 and #482

### LINKED ISSUE #481: Room node not selected when created
### Bug description

When a Room is created, its root node is not selected in the Scene tree. Additionally, the first prop created after installing the plugin doesn't open its folder, nor does it select its node in the tree.

###  Steps to reproduce

1. Create a new project.
2. After completing the setup, create a room.
3. The root node is not selected, but you can still create props.
4. Create a prop.
5. The folder for the prop is created, but its _.tscn_ file is not selected in the **FileSystem**, and the node of the prop is not selected in the tree.

When a room is created, its root node should be selected automatically.
When the first prop in the project is created, its _.tscn_ file should be selected in the **FileSystem**, and its node should be selected in the tree.

 - **OS type and version**: Windows 11
 - **Godot version**: 4.6
 - **Plugin version or branch**: develop

### Additional context

None.

### LINKED ISSUE #482: Polygon editing causes issues if it is not canceled.
### Bug description

When editing the polygon of a **WalkableArea** or a **Clickable**, changing the scene or selecting another node without exiting that editing mode causes usability issues. In the case of changing the scene, the _Output_ displays a large number of errors with the text:

`ERROR: editor/editor_data.cpp:1232 - Condition "!p_node->is_inside_tree()" is true.`

In the case of selecting another node, the Inspector does not display its properties because it is still in polygon editing mode.

###  Steps to reproduce

1. Create or select a **WalkableArea** for a **Room**.
2. Clic the button to edit its polygon.
3. Change to another scene or select another node in the tree.
4. If you change to another scene you'll see the error mentioned above appear several times in the _Output_. If you select another node, the _Inspector_ won't show its properties, and you'll still see the properties of the polygon being edited.

When moving to another scene or selecting another node in the _Scene_ tree, the polygon edition should cancel to avoid usability issues.

 - **OS type and version**: Windows 11
 - **Godot version**: 4.6
 - **Plugin version or branch**: develop

### Additional context

None.

mapedorr:
- Fix #481 : Scan the filesystem after creating objects with custom logic instead of using `PopochiuEditorHelper.filesystem_scanned()` because it was failing due to a script reload.
- Fix #482 : Exit polygon edition when selecting a node in the _Scene_ dock or when changing to another scene or closing the current one.

### DIFF
--- a/addons/popochiu/editor/canvas_editor_menu/popochiu_canvas_editor_menu.gd
+++ b/addons/popochiu/editor/canvas_editor_menu/popochiu_canvas_editor_menu.gd
@@ -34,9 +34,11 @@ func _ready() -> void:
 	btn_interaction_polygon.pressed.connect(_select_interaction_polygon)
 	btn_obstacle_polygon.pressed.connect(_select_obstacle_polygon)
 
-	# Connect to singleton signals
+	# Connect to global signals
 	EditorInterface.get_selection().selection_changed.connect(_on_selection_changed)
 	EditorInterface.get_editor_settings().settings_changed.connect(_on_gizmo_settings_changed)
+	PopochiuEditorHelper.signal_bus.scene_changed.connect(_on_scene_changed)
+	PopochiuEditorHelper.signal_bus.scene_closed.connect(_on_scene_closed)
 
 	_set_toolbar_buttons_color()
 	hide()
@@ -44,7 +46,7 @@ func _ready() -> void:
 
 #endregion
 
-#region Private ####################################################################################
+#region Signals ####################################################################################
 func _toggle_markers_visibility() -> void:
 	PopochiuEditorHelper.signal_bus.gizmo_visibility_changed.emit(
 		PopochiuGizmoPlugin.MARKER_POS,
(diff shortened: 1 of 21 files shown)


addons/popochiu/editor/canvas_editor_menu/popochiu_canvas_editor_menu.gd
```diff
@@ -580,4 +609,37 @@ func _reset_buttons_state() -> void:
...
+func _is_selection_from_scene_tree() -> bool:
+	var mouse_pos := get_viewport().get_mouse_position()
+	var scene_tree_dock := _find_scene_tree_dock()
+	
+	if scene_tree_dock:
+		var rect := scene_tree_dock.get_global_rect()
+		return rect.has_point(mouse_pos)
+	
+	# If we can't find the scene tree dock, assume false (don't cancel)
+	return false
```
stickgrinder, line 623: This is the only thing that smells a bit too much. What if I change the selection by navigating via the keyboard?

Locking the user into an editing mode was a way to avoid frustration with misplaced clicks. Since that's the one and only case we need to cover I'd do the other way around: if the mouse is on the scene editor, then we DO NOT change the node. Other means are OK.

I think this is a bit to hacky and restrictive.

addons/popochiu/editor/popups/create_object/create_object.gd
```diff
@@ -68,7 +68,11 @@ func create() -> void:
 	var created_object := await _create()
 	if not created_object or not is_instance_valid(created_object):
 		return
-	await PopochiuEditorHelper.filesystem_scanned()
+	
+	# Fix #481: Scan the filesystem directly instead of using the PopochiuEditorHelper function
```
stickgrinder, line 72: Fuck...

Issue #487, opened 2026-02-24
Bug: Using a typed Array will break loading.


jmschrack:
### Bug description

If using a Array of a built-in type (i.e. `Array[bool]`),  Popochiu will throw an error when loading.

> Invalid assignment of property or key 'someData' with value of type 'Array' on a base object of type 'Resource (inventory_item_someitemhere_state.gd)'.

popochiu_save_load.gd:197

Changing the type to an untyped `Array` resolved the issue.

###  Steps to reproduce

1. Create an item
2. Add a variable to it's state with a typed array.   i.e. `var some_data: Array[bool]=[true,false]`
3. Run the game, add the item to your inventory.  Save your game.
4. Load the game you saved, see error.

### Expected vs observed behavior
I would expect a typed array of builtins to work given that an untyped array does.

### Environment information (please complete):

 - **OS type and version**: Win11
 - **Godot version**:  4.6.1
 - **Plugin version or branch**:  dev

PR #488, opened 2026-02-24
Added typesafe assignment for arrays when being loaded

jmschrack:
This fixes https://github.com/carenalgas/popochiu/issues/487  by checking if a property is a typed array, and then calling "assign" if it is.

This also fixes a similar bug when loading a room from a save file. it seems the walkable area is not getting deserialized. It's just an array of string. Declaring the interaction_polygon as a `Array[PackedVector2Array]` then using the similar logic in the room readied to assign the array, handles deserialization properly.

### DIFF
--- a/addons/popochiu/engine/objects/walkable_area/popochiu_walkable_area.gd
+++ b/addons/popochiu/engine/objects/walkable_area/popochiu_walkable_area.gd
@@ -14,7 +14,7 @@ extends Node2D
 @export var enabled := true: set = _set_enabled
 ## Stores the outlines to assign to the [b]NavigationRegion2D/NavigationPolygon[/b] child during
 ## runtime. This is used by [PopochiuRoom] to store the info in its [code].tscn[/code].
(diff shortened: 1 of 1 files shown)


addons/popochiu/engine/objects/walkable_area/popochiu_walkable_area.gd
```diff
@@ -14,7 +14,7 @@ extends Node2D
 @export var enabled := true: set = _set_enabled
 ## Stores the outlines to assign to the [b]NavigationRegion2D/NavigationPolygon[/b] child during
 ## runtime. This is used by [PopochiuRoom] to store the info in its [code].tscn[/code].
-@export var interaction_polygon := []
+@export var interaction_polygon :Array[PackedVector2Array]= []
```
stickgrinder:
```suggestion
@export var interaction_polygon:Array[PackedVector2Array] = []
```

Issue #243, opened 2024-06-02
Create a gizmo for polygon editing


stickgrinder:
### Benefit description

Currently, all the Popochiu Objects that depend on a polygon (be it an `InteractionPolygon2D` or a `NavigationRegion2D`), allow the editing of the polygon shape through a button that selects the inner node "behind the scene".

This works, but it is susceptible to a lot of problems. For example:

* It is difficult to manage the selection code and it requires a bit of trickery to jump back and forth between the parent node and the polygon inside it
* If a node has more polygons (to represent "holes", or composite colliders - see #10 ), a single button to edit them all simply doesn't work
* The user has to press a button to edit the polygon explicitly, or even just to see it on the scene, making the interaction annoying.

We need a better solution

### Solution description

As silly as it sounds, we need to implement a new 2D gizmo, using the same strategy we used in #232.

By painting on the viewport we may create a polygon gizmo tool that connects to one or multiple elements inside its node, and sports a tailored interaction.

Most of its behaviors will mimic normal polygon editing, but it may be switched from read-only (see-only) to editing mode, or change more polygons at once. Or even allow the creation of "holes" that are very useful in walkable areas.

### Exclusions

- Cutting holes in a polygon may have to be postponed to a second stage.
- No toolbar for the creation of polygons will be implemented at the moment.

### Implications

This gizmo can be based on GD Extensions, as @drbbgh suggested, but this may be unfeasible if we want to represent more polygons with a single gizmo. We have to check.

### Tasks list

* [ ] Evaluate creating specialized classes in a tree, like
    ```mermaid
        classDiagram
            Gizmo2D <|-- Gizmo2DPos
            Gizmo2D <|-- Gizmo2DAxis
            Gizmo2D <|-- Gizmo2DPoly

            class Gizmo2DPos{
                <<Point in 2D space>>
            }
            class Gizmo2DAxis{
                <<Horizontal or vertical position>>
            }
            class Gizmo2DPoly{
                <<Irregular, composite polygon>>
            }
    ```
* [x] Create a `Gizmo2DPoly` class that holds an array of arrays of vertices. Each array of vertices is a copy of a `CollisionPolygon2D.polygon` or a `NavigationRegion2d.navigation_polygon.get_outline()`. Supporting more polygons will allow _clickables_ to have more than one collision polygon, supporting complex cases.
[...]

PR #489, opened 2026-03-01
Solves #243: custom gizmos for polygon editing

stickgrinder:
Before merging this I want to address the toolbar behavior, that needs more love.


stickgrinder:
@mapedorr I consider this feature complete. It adds native gizmos for polygons, passive gizmos to show every type of polygons on the scene when editing a room, character or clickable (very handy to see where other things are placed while crafting the rooms), a full set of buttons and options to control the appearance and a migration to keep things tidy.

**WARNING**: I didn't review the coding standards and I will have a couple of days off. Be patient. If you see small things, leave a suggestion. If there are many errors, please just tell me to review the coding standards, and I'll do.

I hope this can make its way into develop, I'm pretty happy with this.

stickgrinder:
@mapedorr just to mention that I addressed even the coding standards part of this and I consider this ready for review.

PR #497, opened 2026-03-16
2 fixes (for Touchscreen + walking after a teleport_to_marker)

Afterworks-Studio:
Previously discussed in Discord as contributions.
See topic "Touchscreen" on December 25
See topic "walking issue" after 2.1 migration on March 26


addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -346,7 +346,10 @@ func room_readied(room: PopochiuRoom) -> void:
 			for property in node_dic:
 				if not PopochiuResources.has_property(node, property): continue
 				if node[property] is Array:
-					node[property].assign(node_dic[property])
+					#node[property].assign(node_dic[property])
+				## HACK 
+					if node_dic[property] is PackedVector2Array: node[property].assign(str(node_dic[property]))
+				##
```
stickgrinder, line 352: Please, avoid if at all possible to send PRs with hacks into them. Structural fixes is always preferred.
If unable or unwilling to send one, no problem, it's better to open an issue and explain how you patched your local copy and we'll take that over.

In that case I would just reduce everything to a call to a helper function that handles the various cases gracefully. As a quick example:

```gdscript
# Restores one node property from stored room state while preserving runtime types.
func _restore_node_property(node: Node2D, property: String, value) -> void:
	var current_value = node[property]

	if current_value is Array:
		var destination: Array = current_value

		if value is Array:
			destination.assign(value)
			return

		if value is PackedVector2Array:
			destination.assign([value])
			return

		node[property] = value
		return

	if current_value is PackedVector2Array and value is Array:
		node[property] = PackedVector2Array(value)
		return

	node[property] = value
```

I'd put this into the `Private` region and call this right after the null check.
Feel free to complete my suggestion if you spot any weak point!

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
```diff
@@ -613,7 +613,7 @@ func _translate() -> void:
 func _is_click_or_touch(event: InputEvent) -> bool:
 	if (
 		(event is InputEventMouseButton and not event.double_click)
-		or (event is InputEventScreenTouch and not event.double_tap)
+		#or (event is InputEventScreenTouch and not event.double_tap)  #HACK to avoid double hits with Touch !
```
stickgrinder: Please, reduce the if to a single line and remove unwanted code. Clean, unlittered code is something we strive towards, whenever possible. If you want to make explicit that handling double-tap is redundant, put a comment on top of the 'if' statement.

addons/popochiu/engine/objects/character/popochiu_character.gd
```diff
@@ -1231,7 +1231,7 @@ func reset_buffered_position() -> void:
 
 ## Syncs the buffered position with the current position to avoid conflicts with walking.
 func sync_buffered_position() -> void:
-	_buffered_position = position
+	update_position()
```
stickgrinder: I'm not sure how this is supposed to work. The goal of this function is to override `_buffered_position` with the actual position to consolidate it entirely. If you call `update_position()` you will assign `_buffered_position` to itself every time you it's not null, defeating the purpose of this call.

I may be missing something, can you explain your chain of thoughts here, please?

stickgrinder (changes requested):
Thanks for sending this over @Afterworks-Studio 

I can't take it as it is, but please review my comments.
If you want me to take this over, I'll do.

Thanks

addons/popochiu/engine/interfaces/i_room.gd
Afterworks-Studio: Thanks for this in-deepth reply !
My mistake, I thought I did not send this one (as explained in Discord), 
because I was sure a better solution may exist (and yours is a very good option for sure :-D)   
So please modify that part the way it suits you !

addons/popochiu/engine/objects/clickable/popochiu_clickable.gd
Afterworks-Studio: OK that format was only for the track
I will modify ASAP as requested

addons/popochiu/engine/objects/character/popochiu_character.gd
Afterworks-Studio: Unfortunately, I don't have a full explanation on that...
the chain : update_position() ==> get_buffered_position() 
seems to resolved something with my characters position AND mobility 
where the initial variable setting blocked the mobility of the characters (infinite loop of moon-walking on this last position after a teleport)

stickgrinder: I will try to replicate your issue so I can have a better grasp on what's happening and find the root cause. I'm afraid your change solves the problem by a side-effect.

Afterworks-Studio:
Moonwalk fixed by @stickgrinder 
Try to PR the touchscreen fix alone in a new PR

PR #500, opened 2026-04-04
Fixes #499 - Moonwalk after teleport

### LINKED ISSUE #499: Anti Glide desynchronizes character position after teleport
### Bug description

When Anti Glide Animation is enabled, teleporting a character can leave its visual position and internal buffered movement position out of sync. After a teleport_to_marker or other instant reposition, a subsequent walk_to or walk_to_marker may start pathfinding from a stale buffered position instead of the character’s actual location.

This can cause characters to appear stuck walking in place, never reach the destination, and never emit movement_ended. The issue is more noticeable in edge cases such as paths starting outside walkable areas or when ignore_walkable_areas is enabled.

#### Expected behavior

After any instant reposition, the character’s transform position and buffered movement position should be synchronized so subsequent walking starts from the correct location.

stickgrinder:
This is a proper fix for the bug found by @Afterworks-Studio and @cidneyhamilton on 2.1, where a teleported or otherwhise marker-set character can sometimes walk on the spot instead of moving.

The workaround suggested by @Afterworks-Studio in #497 was incidentally fixing the problem, by creating other corner-cases.

Tests are welcome.


cidneyhamilton:
LGTM in my game.

Issue #505, opened 2026-04-12
Exiting regions won't reset the character's scaling factor


stickgrinder:
### Bug description

Observed during implementation of #432 .

If a character has an interaction polygon larger than the scaling polygon, entering a scaling region scales the character size and speed, but leaving the region doesn't reset it and the scaling remains in place forever.

### Root cause analysis

The character has two collision shapes with different sizes: `ScalingPolygon` (physics shape index 0) and `InteractionPolygon` physics shape index 1). When they exit the region in different physics frames:

- `ScalingPolygon` exits
- `area_shape_exited` fires
- guard passes, but `InteractionPolygon` still overlaps
- `not character in get_overlapping_areas()` is false
- ` _remove_character_scaling_region` gets skipped

Avoiding checking `get_overlapping_areas()` (because it is done at node level, so we get screwed by another polygon still touching the region) seems to fix this problem: basically we don't check for ANY shape to be in the area, just for the (correct) `ScalingPolygon` which is the one to trigger the interaction.

### Additional implications

This also imply that overlapping regions are not managed correctly: the character can have only one active region assigned (that's probably ok).

Also, the `get_overlapping_areas()` check was probably there to make sure that *visually* the character was still scaled even if the ScalingPolygon was a bit off.

To properly fix this bug we probably need to defer a good amount of checks: when the ScalingPolygon leaves the area we mark the character as leaving and we check that every shape has left the area every physics frame, then we reset the flag and that's all.

stickgrinder:
I hardened the logic to check against the correct collider, by using its physics owner, not the position in the scene tree. Now it's more robust.

The problem with overlapping regions is out of scope for this fix, we can work on them if the needs be.

PR #507, opened 2026-04-12
Closes #432: Provides engine API to override character walk speed

### LINKED ISSUE #432: Override Character walk speed
### Benefit description

The character may be made rush or slow down.

### Solution description

* Add `walk_speed_override` variable, that is used unless null or zero
* Add `reset_walk_speed()` which nullifies the above variable

stickgrinder:
Working on that I found some weak spots in the implementation and decided to harden it a bit. One is #505 for which I provided a fix in a specific commit. The other is part of the changes and I'll describe it below.

### Problem

`walk_speed` was playing two roles at once: inspector-configured baseline *and* runtime-scaled effective speed. `update_scale()` was overwriting it every frame whenever a character moved through a scaling region, making it impossible to safely read or override the speed from game scripts. Any value set at runtime would be silently stomped.

### Solution

I separated the concerns cleanly:

- **`walk_speed`** is now a pure inspector property holding the developer's intended baseline. The engine never writes to it at runtime.
- **`_walk_scale_factor`** (private) stores the current perspective scale multiplier, updated by `update_scale()` when the character is inside a `PopochiuRegion`. Defaults to `1.0`.
- **`walk_speed_override`** (public) is a temporary, script-driven speed adjustment. When active (> 0), it replaces `walk_speed` as the base, while `_walk_scale_factor` is still applied on top,  so the override respects game-world perspective.
- **`current_walk_speed`** (public, read-only computed property) is the single source of truth for effective movement speed: `(walk_speed_override if active else walk_speed) * _walk_scale_factor`.
- **`default_walk_speed`** is removed. It was only needed to undo the mutation of `walk_speed`, which no longer happens.

Both `walk_speed` and `walk_speed_override` validate their input: non-positive `walk_speed` is clamped to `1.0` with an error; negative `walk_speed_override` is clamped to `0.0` with a warning.

### Breaking change

`default_walk_speed` is removed. Any game script reading it should be updated to read `walk_speed` directly.

### New public API

- `walk_speed_override: float`: set to override movement speed temporarily; `0.0` = inactive
- `current_walk_speed: float`: read-only; the actual speed used each frame
- `reset_walk_speed()`: convenience method to clear the override (`walk_speed_override = 0.0`)
- `queue_reset_walk_speed()`: queue-chain variant

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -106,8 +106,10 @@ const STANDARD_TALK_ANIMATION = "talk"
 ## You can use this to define which [Texture] to use as avatar for the character when it speaks
 ## using a specific emotion.
 @export var avatars := []: set = set_avatars
-## The speed at which the character will move in pixels per frame.
-@export var walk_speed := 200.0
+## The speed at which the character will move in pixels per frame.[br]
+## Values must be strictly greater than [code]0[/code]. Assigning [code]0[/code] or a negative
+## value will be clamped to [code]1.0[/code] and an error will be logged.
+@export var walk_speed := 200.0: set = set_walk_speed
 ## Whether the character can or not move.
 @export var can_move := true
 ## Whether the character ignores or not walkable areas. If [code]true[/code], the character will move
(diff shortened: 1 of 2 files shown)


stickgrinder:
@mapedorr considering this diff is pretty small and I'm confident about the workings of the feature (it's pretty simple stuff and I tested it), I'm going to merge this. If something goes horribly wrong I'll fix it.

PR #509, opened 2026-04-18
Closes #349 by providing stackable inventory items

### LINKED ISSUE #349: Support for quantities of inventory items
### What's the expected benefit of the feature you are requesting?

I would like to be able to store multiple items of the same type as a quantity of that item.

### Describe the solution you'd like

e.g. if I have 1 coin and I pick up another coin I now have a coin in my inventory with a quantity of 2, not a coin and another coin.
There may be situations where you would like the coins (in this example) as separate inventory items.

### Describe alternatives you've considered

None

stickgrinder:
This PR reworks how "adding to the inventory" works under the hood in a way that allows addition of single quantity items, but also supports "quantity" for items in a specific inventory slots.

For the most part, this is about engine and game devs API, with nothing about visual clues: no counters, no single/multiple items dedicated icons for the inventory items, etc, so it's candidate for some future improvements, but at least, it allows devs to support this much wanted feature out of the box.
Visual clues may be managed with some non-obstrusive strategies in the custom GUI, at the moment.

**IMPORTANT NOTE:** the last commit in this PR introduces an API change by removing the "animate" params from add/remove/replace calls, in a first attempt to solve a strong coupling problem I have found in the Engine/GUI relationship design (see #510). A migration is there to address this, too.

That commit also contains some overall refactoring (my bad, it's a bit too much of a commit) so I can either split the breaking change out of it or we can accept it and go on with a cleaner interface for now.

### DIFF
--- a/addons/popochiu/engine/interfaces/i_inventory.gd
+++ b/addons/popochiu/engine/interfaces/i_inventory.gd
@@ -58,6 +58,10 @@ signal inventory_shown
 ## Emitted when the inventory is requested to hide. [param use_anim] indicates whether the GUI
 ## should use an animation.
 signal inventory_hide_requested(use_anim: bool)
+## Emitted when the quantity of [param item] changes without the item being added to or removed
+## from the inventory (i.e., when stacking or partially removing). [param new_quantity] is the
+## updated count.
+signal item_quantity_updated(item: PopochiuInventoryItem, new_quantity: int)
 
 ## Provides access to the inventory item that is currently selected.
 var active: PopochiuInventoryItem : set = set_active
@@ -91,9 +95,17 @@ func clean_inventory(in_bg := false) -> void:
 		var pii: PopochiuInventoryItem = _item_instances[instance]
 		
 		if not pii.in_inventory: continue
-		if not in_bg: await pii.discard()
 		
-		pii.remove(!in_bg)
+		if in_bg:
+			# refs #349: In background mode, reset quantity_owned directly to avoid signal
+			# emissions and GUI awaits that are not meaningful without a visible inventory.
+			pii.quantity_owned = 0
+		else:
+			# refs #349: Only call discard() here: the old code called both discard() and
+			# remove() separately, which caused item_removed to be emitted twice per item
+			# (once from inside discard() -> remove(), and once from the extra remove() call
+			# after discard()). discard() already calls remove() internally.
+			await pii.discard()
 
 
 ## Shows the inventory for [param time] seconds.
@@ -196,6 +208,13 @@ func is_full() -> bool:
 ## Deselects the [member active] item.
 func deselect_active() -> void:
 	active = null
+
+
+## Returns the number of [param item_name] currently owned by the player.
+## Returns [code]0[/code] if the item is not in the inventory.
+func get_item_quantity(item_name: String) -> int:
+	var i: PopochiuInventoryItem = get_item_instance(item_name)
+	return i.quantity_owned if is_instance_valid(i) else 0
 
 
 #endregion
--- a/addons/popochiu/engine/interfaces/i_inventory.gd
+++ b/addons/popochiu/engine/interfaces/i_inventory.gd
@@ -31,14 +31,12 @@ extends Node
 ## ])
 ## [/codeblock]
 
-## Emitted when [param item] is added to the inventory. [param animate] may be used by the GUI
-## to animate the item entering the inventory.
-signal item_added(item: PopochiuInventoryItem, animate: bool)
+## Emitted when [param item] is added to the inventory.
+signal item_added(item: PopochiuInventoryItem)
 ## Emitted when the [param item] has finished entering the inventory (GUI animation completed).
 signal item_add_done(item: PopochiuInventoryItem)
-## Emitted when [param item] is removed from the inventory. [param animate] may be used by the
-## GUI to animate the item leaving the inventory.
-signal item_removed(item: PopochiuInventoryItem, animate: bool)
+## Emitted when [param item] is removed from the inventory.
+signal item_removed(item: PopochiuInventoryItem)
 ## Emitted when the [param item] has finished leaving the inventory (GUI animation completed).
 signal item_remove_done(item: PopochiuInventoryItem)
 ## Emitted when [param item] is replaced in the inventory by [param new_item]. Useful for
@@ -67,6 +65,9 @@ signal item_quantity_updated(item: PopochiuInventoryItem, new_quantity: int)
 var active: PopochiuInventoryItem : set = set_active
 ## Provides access to the inventory item that was clicked.
 var clicked: PopochiuInventoryItem
+## When [code]true[/code], the inventory is being restored from a save file. GUI components
+## should skip entrance/exit animations during restore.
+var is_restoring := false
 # ---- Used for saving/loading the game ------------------------------------------------------------
 ## [Array] containing instances of the currently held [PopochiuInventoryItem]s.
 var items := []
@@ -87,25 +88,21 @@ func _init() -> void:
 
 #region Public #####################################################################################
 ## Removes all items currently in the inventory. If [param in_bg] is [code]true[/code], items are
-## removed in background without calling [method PopochiuInventoryItem.discard].
+## removed silently without GUI lifecycle (useful during scene transitions).
 func clean_inventory(in_bg := false) -> void:
-	items.clear()
-	
-	for instance in _item_instances:
-		var pii: PopochiuInventoryItem = _item_instances[instance]
-		
-		if not pii.in_inventory: continue
-		
-		if in_bg:
-			# refs #349: In background mode, reset quantity_owned directly to avoid signal
-			# emissions and GUI awaits that are not meaningful without a visible inventory.
+	if in_bg:
+		# refs #349: In background mode, reset quantity_owned directly and clear the items list
+		# to avoid signal emissions and GUI awaits.
+		for instance in _item_instances:
+			var pii: PopochiuInventoryItem = _item_instances[instance]
 			pii.quantity_owned = 0
-		else:
-			# refs #349: Only call discard() here: the old code called both discard() and
-			# remove() separately, which caused item_removed to be emitted twice per item
-			# (once from inside discard() -> remove(), and once from the extra remove() call
-			# after discard()). discard() already calls remove() internally.
-			await pii.discard()
+		items.clear()
+	else:
+		# Remove each item through its full GUI lifecycle.
+		for instance in _item_instances:
+			var pii: PopochiuInventoryItem = _item_instances[instance]
+			if pii.in_inventory:
+				await pii.remove()
 
 
 ## Shows the inventory for [param time] seconds.
--- a/addons/popochiu/engine/interfaces/i_inventory.gd
+++ b/addons/popochiu/engine/interfaces/i_inventory.gd
@@ -108,7 +108,8 @@ func clean_inventory(in_bg := false) -> void:
 
 
 ## Adds [param quantity] of [param item] to the inventory and waits until any GUI transition has
-## finished.
+## finished. Inventory capacity is slot-based: stacked quantities still occupy a single slot and
+## count as [code]1[/code] against the inventory limit.
 func add_item(item: PopochiuInventoryItem, quantity := 1) -> void:
 	if quantity <= 0:
 		PopochiuUtils.print_warning(
@@ -169,13 +170,18 @@ func remove_item(item: PopochiuInventoryItem, quantity: int = 0) -> void:
 
 
 ## Replaces [param item] in the inventory with [param new_item] and waits until the GUI swap has
-## finished.
+## finished. Replacing removes the whole collected quantity of [param item] and adds exactly one
+## quantity of [param new_item].
 func replace_item(item: PopochiuInventoryItem, new_item: PopochiuInventoryItem) -> void:
+	if not _can_replace_item(item, new_item):
+		await get_tree().process_frame
+		return
+
 	_apply_full_removal(item)
 
 	if new_item.quantity_owned == 0:
 		_apply_first_add(new_item, 1)
-	elif new_item.max_quantity > 1 and new_item.quantity_owned < new_item.max_quantity:
+	else:
 		_apply_stack_add(new_item, 1)
 
 	item_replaced.emit(item, new_item)
@@ -303,8 +309,9 @@ func has_item_been_collected(item_name: String) -> bool:
 	return is_instance_valid(i) and i.ever_collected
 
 
-## Returns [code]true[/code] if the inventory has reached the inventory limit
-## configured in the project settings.
+## Returns [code]true[/code] if the inventory has reached the inventory limit configured in the
+## project settings. The limit counts occupied slots, not total owned quantity, so a stacked item
+## still occupies a single slot and counts as [code]1[/code].
 func is_full() -> bool:
 	return (
 		PopochiuUtils.e.settings.inventory_limit > 0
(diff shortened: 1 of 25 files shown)


stickgrinder:
@mapedorr before closing this I may want to rethink the `_do_add` and `_do_remove` naming because it sucks, but I'd like you to provide insights on the GUI-relationship changes if you feel like it.
Thanks

stickgrinder:
@mapedorr I included my solution for #510 in this PR after all. You can check that issue (see the schemas I depicted) to better understand the flow. I restructured the GUI code so compoenents are decoupled and "stupid" and brought the inventory lifecycle logic into the Inventory Interface, leaving the public methods in the item as mere proxies.

This gives each class a specific responsibility. Let me know what you think. At this point I prefer having your feedback than merging right away.

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -83,17 +88,137 @@ func _init() -> void:
...
-	items.clear()
-	
-	for instance in _item_instances:
-		var pii: PopochiuInventoryItem = _item_instances[instance]
-		
-		if not pii.in_inventory: continue
-		if not in_bg: await pii.discard()
-		
-		pii.remove(!in_bg)
+	if in_bg:
```
mapedorr: Maybe in the future we could split this in two functions (`clean_inventory` and `clean_inventory_in_bg`)?

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -83,17 +88,137 @@ func _init() -> void:
...
+
+	if item.quantity_owned == 0:
+		if is_full():
+			PopochiuUtils.print_error("Couldn't add %s. Inventory is full." % item.script_name)
+			await get_tree().process_frame
+			return
+
+		var actual := _get_addable_quantity(item, quantity, item.max_quantity)
+		if actual <= 0:
+			await get_tree().process_frame
```
mapedorr: Should we print something here (same for the `if` statement in line 138) so devs get to know what happened in case an inventory item is not added by this function?

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -83,17 +88,137 @@ func _init() -> void:
...
+			% [quantity, item.script_name]
+		)
+		await get_tree().process_frame
+		return
+
+	var qty_to_remove := quantity if quantity > 0 else item.quantity_owned
+	if qty_to_remove >= item.quantity_owned:
+		_apply_full_removal(item)
+		item_removed.emit(item)
+		await item_remove_done
```
mapedorr, line 184: As we discussed in our meeting, there could be a corner case where an inventory item can "hold" a quantity of items (like a jar full of coins or beans). Devs can manage this by replacing the item (e.g. the jar) when empty, but maybe in the future items could have a property to prevent them to be removed when there's a quantity of 0, or maybe items can have a `min_quantity`. We can talk about this later.

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -83,17 +88,137 @@ func _init() -> void:
...
+	if not _can_replace_item(item, new_item):
+		await get_tree().process_frame
+		return
+
+	_apply_full_removal(item)
+
+	if new_item.quantity_owned == 0:
+		_apply_first_add(new_item, 1)
+	else:
+		_apply_stack_add(new_item, 1)
```
mapedorr, line 205: I'm a bit confused here (maybe because I'm stupid and I can't see the whole thing).

If you wanted to replace an object X with, say, a pack containing 10 cards for mom, would that add one more card to the pack when the new object replaces the previous one?

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -83,17 +88,137 @@ func _init() -> void:
...
+	if value:
+		if item.quantity_owned == 0:
+			_apply_first_add(item, 1)
+			return
+
+		item.ever_collected = true
+		_register_item(item)
+		return
+
+	_apply_full_removal(item)
```
mapedorr: Awesome!

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -211,3 +344,86 @@ func set_active(value: PopochiuInventoryItem) -> void:
...
+func _register_item(item: PopochiuInventoryItem) -> void:
+	if not items.has(item.script_name):
+		items.append(item.script_name)
+
+
+func _apply_first_add(item: PopochiuInventoryItem, quantity: int) -> void:
+	_register_item(item)
+	item.quantity_owned = quantity
+	item.ever_collected = true
+	item._on_added_to_inventory()
```
mapedorr: Calling a method from outside an object should make use of a public method: `item.on_added_to_inventory()`.

addons/popochiu/engine/interfaces/i_inventory.gd
```diff
@@ -211,3 +344,86 @@ func set_active(value: PopochiuInventoryItem) -> void:
...
+
+
+func _apply_stack_add(item: PopochiuInventoryItem, quantity: int) -> void:
+	var old_qty := item.quantity_owned
+	item.quantity_owned += quantity
+	item._on_quantity_changed(old_qty, item.quantity_owned)
```
mapedorr: Same for this one: `item.on_quantity_changed(...)`. It will affect line 425 too.

addons/popochiu/engine/objects/gui/templates/9_verb/9_verb_gui.gd
```diff
@@ -16,6 +16,7 @@ var _return_to_walk_to := false
 ## Used to access the [b]9VerbPanel[/b] component (the one at the bottom containing the verbs,
 ## the inventory, and the button to open the [b]9VerbSettingsPopup[/b].
 @onready var _9_verb_panel: Control = %"9VerbPanel"
+@onready var _inventory_grid: PopochiuInventoryGrid = _9_verb_panel.get_node("%9VerbInventoryGrid")
```
mapedorr: I could be wrong, but accessing the **9VerbInventoryGrid** node shouldn't require looking for it inside the `_9_verb_panel` node.

```suggestion
@onready var _inventory_grid: PopochiuInventoryGrid = %9VerbInventoryGrid
```

addons/popochiu/engine/objects/gui/templates/sierra/sierra_gui.gd
```diff
@@ -11,6 +11,9 @@ extends PopochiuGraphicInterface
 @onready var sierra_bar: Control = %SierraBar
 @onready var sierra_menu: Control = %SierraMenu
 @onready var sierra_inventory_popup: Control = %SierraInventoryPopup
+@onready var _inventory_grid: PopochiuInventoryGrid = sierra_inventory_popup.get_node(
+	"%SierraInventoryGrid"
+)
```
mapedorr: Same as the previous comment.

```suggestion
@onready var _inventory_grid: PopochiuInventoryGrid = 
	"%SierraInventoryGrid"
```

addons/popochiu/engine/objects/gui/popochiu_gui.gd
```diff
@@ -147,6 +150,30 @@ func _on_inventory_item_selected(item: PopochiuInventoryItem) -> void:
 	pass
 
 
+## Called when [param item] is added to the inventory.[br]
+## This hook is awaited by the base GUI routing layer, which emits
+## [signal PopochiuIInventory.item_add_done] automatically after the hook returns. Overrides must
+## not emit the signal or call [method G.block].
+func _on_item_added(item: PopochiuInventoryItem) -> void:
```
mapedorr: To prevent warnings of non used parameters.

```suggestion
func _on_item_added(_item: PopochiuInventoryItem) -> void:
```

addons/popochiu/engine/objects/gui/popochiu_gui.gd
```diff
@@ -147,6 +150,30 @@ func _on_inventory_item_selected(item: PopochiuInventoryItem) -> void:
...
+	pass
+
+
+## Called when [param item] is removed from the inventory.[br]
+## This hook is awaited by the base GUI routing layer, which emits
+## [signal PopochiuIInventory.item_remove_done] automatically after the hook returns. Overrides
+## must not emit the signal or call [method G.block].
+func _on_item_removed(item: PopochiuInventoryItem) -> void:
```
mapedorr:
```suggestion
func _on_item_removed(_item: PopochiuInventoryItem) -> void:
```

addons/popochiu/engine/objects/gui/popochiu_gui.gd
```diff
@@ -147,6 +150,30 @@ func _on_inventory_item_selected(item: PopochiuInventoryItem) -> void:
...
+	pass
+
+
+## Called when [param item] is replaced in the inventory by [param new_item].[br]
+## This hook is awaited by the base GUI routing layer, which emits
+## [signal PopochiuIInventory.item_replace_done] automatically after the hook returns. Overrides
+## must not emit the signal or call [method G.block].
+func _on_item_replaced(item: PopochiuInventoryItem, new_item: PopochiuInventoryItem) -> void:
```
mapedorr:
```suggestion
func _on_item_replaced(_item: PopochiuInventoryItem, _new_item: PopochiuInventoryItem) -> void:
```

addons/popochiu/engine/objects/inventory_item/popochiu_inventory_item.gd
```diff
@@ -22,10 +22,26 @@ signal unselected
...
+# @deprecated: Use [member quantity_owned] instead. Kept for backward compatibility
+#
+## Reading returns [code]true[/code] if [member quantity_owned] is greater than [code]0[/code].[br]
+## Setting [code]true[/code] ensures [member quantity_owned] is at least [code]1[/code].[br]
+## Setting [code]false[/code] sets [member quantity_owned] to [code]0[/code].[br]
+## [b]Warning:[/b] Unlike [method add] and [method remove], this setter does not emit inventory
+## signals and does not update the inventory GUI. Use [method add] and [method remove] instead.
+var in_inventory: bool :
+	get: return quantity_owned > 0
+	set(value): _set_in_inventory(value)
```
mapedorr: Could be to keep the format we've been using in other setter methods.

```suggestion
	set = set_in_inventory
```

addons/popochiu/engine/objects/inventory_item/popochiu_inventory_item.gd
```diff
@@ -372,6 +365,17 @@ func get_description() -> String:
 #endregion
 
 #region Private ####################################################################################
+# Deprecated compatibility setter. This keeps the old silent state-toggling semantics, but new
+# code should use add()/remove() for gameplay flows or PopochiuIInventory.register_existing_item()
+# for GUI bootstrap registration.
+func _set_in_inventory(value: bool) -> void:
```
mapedorr: We can keep this as it is since the method will be removed in the future(?), but to follow our coding standards, this shouldn't have an underscore at the beginning, and should go inside the _SetGet_ region.

```suggestion
func set_in_inventory(value: bool) -> void:
```

mapedorr (changes requested):
I left a couple of change suggestions to follow our standards, and some questions and open comments.

So far, I liked the new approach for handling quantities and GUI feedback.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: I didn't notice this and makes sense! I'll do it in this PR since it is the kind of stuff that's pertaining to the same design logic.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: Makes perfect sense! I'll do.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: The scenario is a logic where an object is swapped for another. If the "other" object is alerady in the inventory and a quantity can be added, we don't want to trigger a "first add", but a "stack to what you have".

I kept this for coherence with the former logic, but let me state this:

1. Having a swap function may be useful but it's not that impacting, just like "grab" on a "linked item", it may do more harm than good at times
2. The whole "first time you play an animation", then no more comes from the behavior of the simpleclick gui and this code currently work with this legacy.

I'd like to discuss further decoupling of the whole GUI, but so far I kept things as they were.
This addresses the corner case I explained above.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: Right! I would leave it as it is and provide a proxy method to preserve the public interface (how I wish gdscript wasn't inspired by python!!!)

addons/popochiu/engine/objects/gui/templates/9_verb/9_verb_gui.gd
stickgrinder: Ah Jesus Christ, this was an error introduced by AI and I forgot to clear it. I noticed it already, thanks for pointing that out!

addons/popochiu/engine/objects/inventory_item/popochiu_inventory_item.gd
stickgrinder: I always struggle remembering setter and getter formats!  Thanks

addons/popochiu/engine/objects/inventory_item/popochiu_inventory_item.gd
stickgrinder: Sure, I don't know why I changed it, but it should have been one of my first commits. I ended up changing a good amount of stuff along the way.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: I also restructured the function a bit so it's more readable. Removed a useless test for a corner case that can never happen (due to validation on the other class).

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: If I remember correctly, we ended up deciding that this can be managed by the game script by swapping the "full jar" with an "empty jar" once you run out of beans in the jar.

I had exactly this situation in PITB and we managed it that way. It worked pretty well.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: I hope I managed to clarify this. Closing for now.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: I added the explicit function, that's way better than a feature flag param, but I think this has to be managed in a totally different way. See #510

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: Uhm... wait, sorry. That's not private, that's a virtual. Technically that one is a legit call, then.
Thanks again to gdscript for the sane coding conventions.

I added a comment to the code. Let me know if you want me to change this.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: See above. Closing this, we can decide there.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: I'll give this for granted.

addons/popochiu/engine/interfaces/i_inventory.gd
stickgrinder: We agreed that this is a naming-convention problem of Godot that we'll address systemically in Popochiu once we settle on a final style.

For now we'll live these untoched so they stay consistent with the rest of the code.

Issue #504, opened 2026-04-08
Touchscreen feature works differently, depending on the target build (project settings)


Afterworks-Studio:
### Conditions with no issue (Touchscreen working as expected)

WIN10 : disable "emulate Touch from mouse" in the project settings
Android : On the contrary, setting should be enabled
Web : the setting should be enabled (and it works on both WIN10 or Android device using Chrome !)
iOS : not tested

### Bug description
In the opposite conditions, click interactions don't trigger totally (nothing happens) or partially (moving is possible but no triggers on hotspots etc) --> depending on the target OS 

### Environment information :

 - **Godot version**: 4.6.1
 - **Plugin version or branch**:  develop branch ; last repo update 

### Additional context

During investigation I put a print(index) in popochiu_utils.gd > static func get_click_or_touch_index
An extract of the log result (tested on WIN10) :
169
1
0
0
0
0
170

1 = a trigger index 
0 = no touch or drag from touchscreen
higher numbers = the sum of triggered actions or screentouch, including those from previous sessions of play test ! 
(so var index alternatively -with touchscreen, and NOT with mouse clics in between- is updated with value not only corresponding on the strict index from multi-touch (fingers from 1 to 10 for instance)

stickgrinder:
Thanks for reporting this. I will look into that asap.

stickgrinder:
@Afterworks-Studio PR #512 should normalize the behavior addressing the root cause of the bug. Can you please test this and see if it works?

Some verification steps you want to check:

* Desktop, emulation OFF: mouse click on hotspot fires exactly once; double-click detects correctly
* Desktop, emulation ON: mouse click fires exactly once (no double-firing from synthetic touch); floor click-to-walk works
* Android/Web (real touchscreen): tap on hotspot invokes left-click handler; tap on floor walks player; inventory item tap works
* Desktop, emulation ON: tapping dialog_text, dialog_menu, popups all respond correctly to touch

Afterworks-Studio:
Hi !

Quick feedback from my side after having picked your PR #512 [N.B. :  no
"mouse" regression with any settings] :

→ Clic emulation ON
- Android Web Touch (html5) :
     texts are cut off directly for the next sentence (no full sentence print before next clic) 
clics in inventory do not work
		→ possible common bug : silent double “clic” somewhere

- Android Touch (apk) : 
Same bugs as Web Touch

- PC Touch : 
[...]

PR #512, opened 2026-05-09
refs #504: first finger tap event now normalized to mouse click

stickgrinder:
Also refactored PopochiuClickable local code to rely on PopochiuUtils. General improvement.


mapedorr (approved):
Great!

Issue #521, opened 2026-05-21
Regions can block transpassing


stickgrinder:
Regions can be made walkable/non-walkable to block characters passage on an otherwise fully active walkable area.

This can be interesting in those case where a prop obstacle (or other contextual element) blocks the passage in a zone of a location until it's removed).

* [ ] Add a `walkable` public and exposed variable to regions. It can be set via scripting of course and gets properly saved.
* [ ] Add a "Ignores blocking regions" flag to characters (tooltip hint is like "Can walk on regions with walkable property set to false"), alongside the other region tinting / scaling ignore flags, right after "Ignore obstacles".
* [ ] Characters can't enter a non-walkable region (unless they ignore the blocking ones).
* [ ] If a character is inside a blocked region, it can exit it, but it can't enter it.

### Behavior

When a character tries to enter a blocked region, its movement should be interrupted and a specific signal should fire so the game script can react to this.

We may also want to implement a hook in the character template to manage that signal, receiving the identifier of the region that blocked the character. This way the user can include logic in there.

PR #523, opened 2026-05-22
refs #521: Implement blocking regions

stickgrinder:
### Region Blocking for Character Navigation

Characters are now prevented from entering regions marked as non-walkable.
To achieve this, the navigation path is trimmed before movement: if a segment would cross into a blocking region, the path is cut at the nearest intersection, and the character stops just before the boundary.
The nearest intersection is the one who wins the trim.
In addition, signals are emitted when a character is blocked by a region, so the engine can react.

A first implementation was testing the areas collisions, but this was creating a problem: since the logic was triggered by the character **entering** a blocked region, clicking again on the destination was unblocking the path since the character was already inside the region.

A `BLOCKING_REGION_PULLBACK` constant (current value 2) representing a distance in pixels to trim the segment from the area border, to avoid ambiguous boundary cases (if the character stops on the edge of the region, it is still considered inside, triggering the case above).

Performance impact is negligible: the check is only performed when a new path is set, not during walk, and the number of regions is typically small in a room. In a room with three regions (I consider this pretty crowded compare to average), we can get ~15-20 aritmethic operations per click, which is unmeasurable on modern hardware.

### DIFF
--- a/addons/popochiu/engine/objects/character/popochiu_character.gd
+++ b/addons/popochiu/engine/objects/character/popochiu_character.gd
@@ -51,6 +51,9 @@ signal obstacle_state_changed(character: PopochiuCharacter)
 ## Emitted during movement when the character's position changes.
 ## Only emitted while the character is moving and the position has actually changed from the last emission.
 signal position_updated(character: PopochiuCharacter, current_position: Vector2)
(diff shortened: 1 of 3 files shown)


mapedorr (approved):
Nice and clean!

Issue #42, opened 2023-03-26
Add support for localization


mapedorr:
We should add some way to help developers add localization to their game without changing the syntax of dialog lines wrote in scripts and object properties (i.e. prop descriptions, dialog options).

We can achieve this in two steps if necessary:

1. Integrate [Localization using gettext](https://docs.godotengine.org/en/stable/tutorials/i18n/localization_using_gettext.html).
2. Add support for CSV.

An idea that came to the table is to have a component (could be something by room, or a new tab in Popochiu's dock) that will automatically list all sensitive texts to be translated into several languages with input fields where devs can put the desided localized text.

But a first implementation could be just something that allows devs to use [Poedit](https://poedit.net) to transalate those texts. For this approach, Popochiu should take into account some form of ID assignation for each of those texts sensitive to be translated. Could be something grouped by room, character, inventory item, and dialog (since they are the main objects of the plugin).

I'll put this here just to save this idea:

**RoomHouse.gd**
```gdscript
func _on_room_transition_finished() -> void:
	await C.player.say("Here we go again") # 101_rtf_player_01
	await C.player.say("This time I will not listen to the magician") # 101_ortf_player_02
	await C.player.say("I am tired of playing this game over and over again") # 101_ortf_player_03
```
> The comments in the previous line could be something generated by Popochiu to identify those dialogue lines internally. And those would be the ones that could be used by the runtime localization system.

mapedorr:
I'm gonna leave some comments here while I take a look to a project that is using **gettext** to add localization to a game made in Popochiu.

## Things Popochiu might do

- [ ] Provide a button that will create a `.pot` file with all the `description` values for Props, Hotspots and Inventory items.
- [ ] Provide a button that will create a `.pot` file with all the dialogue lines of the game. This one should take into accoun any **PopochiuDialog** and those lines triggered by `PopochiuCharacter.say` and `E.queue([])` and `E.cutscene([])`.

stickgrinder:
@idbrii I came to the same conclusion by myself and I opened a proposal [here](https://github.com/godotengine/godot-proposals/issues/14908).

Anyone who wants to chime in is welcome. The more traction it gets, the higher the chance to have it included someday.

PR #528, opened 2026-05-27
refs #42: Automated i18n strings extraction from custom Popochiu functions

stickgrinder:


stickgrinder:
@mapedorr I'm ready to explain the overall structure of this work with you, if you want.

addons/popochiu/editor/plugins/popochiu_translation_parser_plugin.gd
```diff
@@ -0,0 +1,800 @@
...
+	_append_extra_names(names, PopochiuConfig.get_translation_extra_plural_function_names())
+	return names
+
+
+func _append_extra_names(names: PackedStringArray, extra: String) -> void:
+	if extra.is_empty():
+		return
+	for n in extra.split(",", false):
+		var trimmed := n.strip_edges()
+		if not PopochiuEditorHelper._is_valid_function_name(trimmed):
```
mapedorr: Functions called from outside a script should expose a public method instead of accessing the private one directly:

```suggestion
		if not PopochiuEditorHelper.is_valid_function_name(trimmed):
```

addons/popochiu/editor/plugins/popochiu_translation_parser_plugin.gd
```diff
@@ -0,0 +1,800 @@
...
+			or _try_singular_match()
+			or _try_text_assignment()
+		)
+		if not matched:
+			_check_unused_modifiers()
+
+	_parse_lines = PackedStringArray()
+	var result := _parse_result
+	_parse_result = []
+	return result
```
mapedorr, line 361: Just out of curiosity: This is to duplicate the array?

mapedorr (changes requested):
Amazing work. Most of the logic in the Translator Parser Plugins looks like black magic to me (specially trying to understand RegEx just from reading them), but I understand the purpose of the parsing.

I just left a picky change suggestion, and a question about a couple of lines (just to understand what's the idea behind them).

addons/popochiu/editor/plugins/popochiu_translation_parser_plugin.gd
stickgrinder: I'm the usual idiot!
I totally overlooked this silliness.

addons/popochiu/editor/plugins/popochiu_translation_parser_plugin.gd
stickgrinder: What I'm trying to do here is returning a stateful member value, while resetting the member itself.

It's not elegant, I admit. `_parse_result` is a class property, I need it clean but I also need to return its value before cleaning it, so I need to save the value, clean it out and return.

PR #530, opened 2026-05-31
refs #360: Update calculations and change visible_characters_behavior…

### LINKED ISSUE #360: The dialog resize function does not work in the 2 button interface

mapedorr:
### Update calculations and change `visible_characters_behavior` in all **RichTextLabel** nodes

The new calculations are for the **DialogCaption** component, which wasn't placing correctly, and which was also re-writing the assignation of the color for the dialogue line (which wasn't necessary since it was done be its parent class, **DialogText**).

Additionally, now all the **RichTextLabel** nodes in **DialogText** components use a Characters After Shaping behavior so the text doesn't move or change its row while being tweened.

This fixes #360 .

### Other changes/updates

- Create the lacking **DialogCaptionHighRes** component.

### DIFF
--- a/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
+++ b/addons/popochiu/engine/objects/gui/components/dialog_text/dialog_caption/dialog_caption.gd
@@ -4,15 +4,20 @@ extends PopochiuDialogText
 
 #region Private ####################################################################################
 func _modify_size(msg: String, _target_position: Vector2) -> void:
-	var size := await _calculate_size(msg)
+	var calc_size := await _calculate_size(msg)
 	
 	# Define size and position (before calculating overflow)
-	rich_text_label.size.y = size.y
-	rich_text_label.position.y = get_meta(DFLT_POSITION).y - (size.y - get_meta(DFLT_SIZE).y)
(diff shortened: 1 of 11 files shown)


stickgrinder (approved):
Go for it!

Issue #417, opened 2025-07-20
Discussion: Popochiu Engine Public API improvements


stickgrinder:
This issue collects a comprehensive analysis of the current Popochiu public API, highlighting missing features, inconsistencies, and areas for improvement, comparing it to other engines (mainly PowerQuest and AGS), in the light of the development of preview games with these engines.

It includes a checklist of subsystem APIs, open questions, and a list of useful game script hooks that should be considered for future development.
The goal is to make Popochiu more flexible, consistent, and developer-friendly, especially for advanced adventure game workflows.

---

## General API Design Principles

- **Consistency**: All subsystems (Character, Prop, Inventory Item, etc.) should expose animation and override functions with a consistent naming and usage pattern.
- **Property vs Method**: Prefer properties for state inspection (e.g. `is_moving`, `is_animating`) and methods for actions (e.g. `fade_in()`, `move_to()`).
- **Overrides**: Use properties to set overrides (e.g. `animation_override`, `position_override`) and provide reset methods to clear them.
- **Read-only State**: Properties that reflect interaction history (e.g. `first_use`, `ever_collected`, `visited`) should be read-only.
- **Nullable Returns**: Functions that query objects under the mouse or in the scene (e.g. `get_clickable_under_mouse()`) should return `null` if nothing is found.
- **Extensibility**: Prefer flexible models for interaction tracking (e.g. `ever_invoked(command)`, `count_invoked(command)`), allowing for future expansion.
- **Editor Integration**: Properties that affect runtime behavior (e.g. `walkable`, `hide_when_blocking`, `vertical_resolution`) should be editable in the inspector and modifiable at runtime.

## Notes

- Animation triggers (add, remove, clear, wait, etc.) from PowerQuest may not map directly to Godot/Popochiu; review their necessity before implementing.
- Cutscene handling and background conversations require special attention for dev-friendliness and non-blocking logic.
- Inventory management per character is not yet implemented; consider API needs before starting.
- Cursor, dialog, and engine subsystems should provide override/reset patterns for all temporary states.
- Room and region APIs should allow for runtime changes and querying, but keep the interface minimal and predictable.
- Settings (e.g. audio volumes, cursor lock) should be exposed via dedicated objects, not through engine singletons.

---

### Clickable

- Use flexible interaction tracking: `ever_invoked(command)`, `first_invoked(command)`, `count_invoked(command)`.
- `enable()`/`disable()` should affect both visibility and interactivity; add `is_enabled()` getter.
[...]

Issue #219, opened 2024-04-23
Save and Load game bug


mapedorr:
### Bug description

Based on a message by @anthonyirwin82 in Discord's 🐞 **| bugs** channel.

- If you have a game you save after state.visited_first_time has run then load the game again it will run the code despite already being visited.
- When you load a game most users would expect the menu to disappear and be in the loaded game.
- When you save a game most users would expect the menu to disappear and be back in the game.

https://github.com/carenalgas/popochiu/assets/4536477/43c65be9-c91c-49fd-bab3-7ee9b316bc88

_Video provided by Anthony._

###  Steps to reproduce

1. Make the character say something when the room is visited for the first time.
2. Save the game.
3. Load the game.
4. The character will say the lines despite the game was saved after it says them (after the room was already visited).

### Expected vs observed behavior

The `visited_first_time` state property should be stored when the game is saved.

### Environment information (please complete):

 - **OS type and version**: Arch Linux
 - **Godot version**: 4.2.2
 - **Plugin version or branch**: Beta 3

PR #538, opened 2026-06-20
refs #219: proposed a more robust solution

stickgrinder:
@mapedorr this PR proposes a more robust solution to the one-line fix in #219 which seems to me a little too violent.

The core problem I see is that `visited_first_time` was being altered *after* `_on_room_transition_finished()` to mask a bug where that hook fires even when restoring from a save. This corrupts the game-state semantic of `visited_first_time` a room that *was* visited once gets reloaded as not visited or visited more than once, which is wrong conceptually and can break any game code that reads the property outside the hook.

What I changed and why:

1. **`PopochiuRoomData`: `visited` and `visited_first_time` became read-only getters**  
   `visited` → `visited_times > 0`  
   `visited_first_time` → `visited_times == 1`  
   `visited_times` is now the single persisted source of truth. This eliminates the risk of stale/contradictory values when saving and loading and goes into the direction I thought when I noted them in #417.

2. **i_room.gd: `_on_room_transition_finished()` is now skipped when restoring from a save.**  
   I would argue that hook was never designed for savegame restoration: its purpose is to react to *entering* the room by processing user-visible events (cutscenes, starting sounds, etc). Replaying that on load may cause bugs (e.g. re-triggering first-visit cutscenes, re-positioning characters).

3. **`PopochiuRoom`: new `_on_restore_from_savegame()` virtual**
   This new virtual is called instead, giving game authors an explicit, intentional hook for load-specific logic (ambient audio, signal reconnection, resumed events). Documented as *not* a replacement for `_on_room_entered()` (which still runs on both fresh and loaded entries for setup).

4. **Room template updated**: new rooms get the stubbed new virtual with Migration 16.

**Known caveat**  
Any existing game script that assigns to `state.visited` or `state.visited_first_time` will error since they're now read-only. The migration prints a grep hint; the author must audit their own codebase.

I may have completely misinterpreted how the system is supposed to work, so I kindly ask you to review my assumptions in addition to the implementation.

Thanks!

### DIFF
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -380,7 +380,7 @@ func room_readied(room: PopochiuRoom) -> void:
 
 	PopochiuUtils.e.in_room = true
 
-	var _was_loaded := not PopochiuUtils.e.loaded_game.is_empty()
+	var was_loaded := not PopochiuUtils.e.loaded_game.is_empty()
 
 	if PopochiuUtils.e.loaded_game:
 		PopochiuUtils.e.game_loaded.emit(PopochiuUtils.e.loaded_game)
@@ -391,7 +391,7 @@ func room_readied(room: PopochiuRoom) -> void:
 	# This enables the room to listen input events
 	current.is_current = true
 
-	if _was_loaded:
+	if was_loaded:
 		await current._on_restore_from_savegame()
 	else:
 		await current._on_room_transition_finished()
--- a/addons/popochiu/engine/interfaces/i_room.gd
+++ b/addons/popochiu/engine/interfaces/i_room.gd
@@ -241,9 +241,7 @@ func room_readied(room: PopochiuRoom) -> void:
 	if PopochiuUtils.e.loaded_game:
 		PopochiuUtils.c.player = PopochiuUtils.c.get_character(PopochiuUtils.e.loaded_game.player.id)
 	else:
-		current.state.visited = true
 		current.state.visited_times += 1
-		current.state.visited_first_time = current.state.visited_times == 1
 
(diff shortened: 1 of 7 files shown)


addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -394,6 +380,8 @@ func room_readied(room: PopochiuRoom) -> void:
 
 	PopochiuUtils.e.in_room = true
 
+	var _was_loaded := not PopochiuUtils.e.loaded_game.is_empty()
```
mapedorr: Remove the underscore since Godot uses that to avoid warnings when a variable inside a function is declared but not used.

```suggestion
	var was_loaded := not PopochiuUtils.e.loaded_game.is_empty()
```

addons/popochiu/engine/interfaces/i_room.gd
```diff
@@ -402,10 +390,11 @@ func room_readied(room: PopochiuRoom) -> void:
 
 	# This enables the room to listen input events
 	current.is_current = true
-	await current._on_room_transition_finished()
 
-	# Fix #219: Update visited_first_time state once _on_room_transition_finished() finishes
-	current.state.visited_first_time = false
+	if _was_loaded:
```
mapedorr:
```suggestion
	if was_loaded:
```

mapedorr (changes requested):
As for the logic, I think this approach is better. It simplifies the code and improves its readability.

I only left one suggested change—please apply it before merging. The comment is based on following Godot's style guide.

stickgrinder:
> As for the logic, I think this approach is better. It simplifies the code and improves its readability.
> 
> I only left one suggested change—please apply it before merging. The comment is based on following Godot's style guide.

Done! Thanks a lot for noticing!

PR #548, opened 2026-07-30
Ensure PC spawns inside walkable area if available

bonki:


stickgrinder:
Nice addition, even thought in a real game the developer is expected to use a marker and some logic in `_on_room_entered` to place the character in the right starting position depending on the context. That's why the code was pretty agnostic.

Before merging this I would review the logic a bit: now you are placing the character in the room center and re-placing it inside the WA if there is an active one. Having an active WA is very likely for any non-stubbed room, so I would avoid the double positioning (as trivial as it is) and check for this case first. If you find none, place the character by using the camera rect.

bonki:
I understand, I implemented this mainly because I find it annoying during testing/prototyping that rooms without a PC spawn the PC in the middle of the screen and once you click you walk through the air to the WA (if one exists). The fallback at center/room bottom is really just a nicer starting point (before it was screen center) because WAs tend to be closer to the bottom, even if they might not exist yet. Later I decided it would also make sense to use this even in the default case of an existing PC in the room (the repositioning always happens now, as you noted) so you would not need to place the PC as carefully, but thinking about it now this also moves the PC *away* from its intended position to the edge of the WA if its already inside a WA, which isn't great and not what a dev would expect. I'll update this (in a couple of weeks, I'm on the go right now) accordingly. Thanks for the feedback!

stickgrinder:
Perfect, I'd be happy to receive your contribution!
Cheers and stay safe

stickgrinder:
Hi @bonki! Any idea about that? Do you want to work on this more or should I close it?

As convenient as it sounds, the main case is solved by properly placing the character at the right position programmatically, so, unless properly addressed as in the previous comment, I would avoid bringing it more logic at the moment.

Let me know. Thanks

Issue #115, opened 2023-12-07
Scrolling rooms support parallax


stickgrinder:
### Benefit description

Another common feature in adventure games, if a room larger than the viewport has many layers (usually true for exteriors or rooms with windows facing exteriors) is for those layers to side-scroll at different speeds (parallax).

### Solution description

Props in rooms may be assigned an integer positive or negative number, mildly representing the distance from the camera, being 0 (zero) a mid-distance. Let's call this number _z-index_.

This number may be an absolute or relative one (that's to say a certain number like 9 always means that the prop will scroll at a certain speed, or it may depend on the extremes on a per-room basis).

### Exclusions

Every item with no z-index should default to zero.

### Implications

None obvious ones.

PR #549, opened 2026-07-30
refs #115: Add parallax support

bonki:
I well-intentionedly chose to ignore the proposed implementation of a simple `z-index` and implemented it similarly to what PowerQuest does, having two settings called `parallax_anchor` (PowerQuest calls this `parallax_alignment`) and `parallax_scroll_scale` (PowerQuest calls this `parallax_depth`).

`parallax_anchor` is responsible for configuring "at this room's relative camera position, position the prop exactly here" - this makes it simple to position props where they appear in the editor for the camera extremes/room bounds.

`parallax_scroll_scale` configures the depth as a factor of the camera movement. I went with this name because that's what `Parallax2D` uses as it describes the relationship with the camera better (`0.5` == "prop moves 50% of camera movement").

### DIFF
--- a/addons/popochiu/engine/objects/prop/popochiu_prop.gd
+++ b/addons/popochiu/engine/objects/prop/popochiu_prop.gd
@@ -50,6 +50,40 @@ signal obstacle_state_changed(prop: PopochiuProp)
 @export_range(0.0, 1.0) var alpha: float = 1.0: set = set_alpha
 ## Total frames available the texture image has. [code](frames * vframes)[/code]
 var total_frames: get = get_total_frames
+## A per-axis scroll-speed multiplier, where [code]1[/code] means "100% camera speed".
+## Positive values move with the camera, negative values move opposite of the camera.[br]
+## [br]
+## Examples:[br]
+## [br]
+##   - [code]   [0, 0][/code]: Disables parallax scrolling on both axes.[br]
+##   - [code][-0.1, 0][/code]: Moves the prop away from the camera on the x-axis at 10% camera speed.
+@export var parallax_scroll_scale: Vector2 = Vector2.ZERO
+## Specifies the parallax prop's "neutral" position, i.e. at which normalized camera position the
+## prop renders at its original position as seen in the editor, where [code]-1[/code] means
+## "100% left/top", [code]0[/code] means "center" and [code]1[/code] means "100% right/bottom".
+## Deviations from the given position are rendered at a shifted position in accordance with
+## [member parallax_scroll_scale].[br]
+## [br]
+## This allows for deterministically anchoring a prop in its editor position ("what you see is what
+## you get"), irrespective of how big the actual overall texture is. This is especially useful for
+## room-sized props coming from an Aseprite-imported layer where the visible pixels only occupy a
+## part of the room.[br]
+## [br]
+## Examples:[br]
+## [br]
+##   - [code][ -1, 0][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             at its leftmost (x-axis) center (y-axis) position."[br]
+##   - [code][0.5, 0][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             halfway between the room center and its rightmost (x-axis) center
+##                             (y-axis) position."[br]
+##   - [code]  [1, 1][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             at its topmost (x-axis) rightmost (y-axis) position."[br]
+@export var parallax_anchor: Vector2 = Vector2.ZERO
+
+var _original_position: Vector2
+@onready var _parallax_enabled: bool = PopochiuUtils.e.settings.parallax_scrolling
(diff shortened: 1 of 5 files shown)


bonki:
Side note: Popochiu is currently somewhat broken with stretch aspect settings `keep_height`/`keep_width` as all the GUI etc. calculations are based on the root viewport which in this case can grow larger than the room. Fixing this was out-of-scope for adding parallax support, but I introduced `E.get_visible_room_rect()` so that at least the parallax props are not all over the place when resizing the window with said aspect settings while all other current limitations persist.

This *could* be used to somewhat mitigate the overall issue by carefully replacing the right calls to `get_viewport().get_visible_rect()` with `get_visible_room_rect()`, having a dedicated `get_mouse_position()` based on `get_visible_room_rect()` and having `E.width`/`E.height` et al be based on `get_visible_room_rect()` as well - I experimented a bit with this and it sort of works, the main problem is that (I assume) due to the GUI elements having a different rendering pipeline they don't always perfectly align with the room props after stretch (the offsets are subpixel-wide). This wouldn't be a huge issue but is quite apparent with e.g. the inventory bar, which is expected to be the same width as the "visible game area" and those subpixel offsets are annoying. I am new to Godot but the canonical way of fixing this, from my current understanding, would be to have a `CenterContainer`/`SubViewportContainer`/`SubViewport`-layout using `size_2d_override` on the `SubViewport` (and update the correct properties when resizing) and have the world and GUI elements share the same `SubViewport`. This works well in my standalone test but given the Popochiu architecture this is not really feasible. Nonetheless, I put `get_visible_room_rect()` into the core engine because it might prove useful in other places and maybe the hackish approach I described above might be worth considering as an interim solution even if it's not perfect.

addons/popochiu/engine/objects/prop/popochiu_prop.gd
```diff
@@ -50,6 +50,40 @@ signal obstacle_state_changed(prop: PopochiuProp)
...
+## Examples:[br]
+## [br]
+##   - [code][ -1, 0][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             at its leftmost (x-axis) center (y-axis) position."[br]
+##   - [code][0.5, 0][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             halfway between the room center and its rightmost (x-axis) center
+##                             (y-axis) position."[br]
+##   - [code]  [1, 1][/code]: "Render the prop at its initial (editor) position when the camera is
+##                             at its topmost (x-axis) rightmost (y-axis) position."[br]
+@export var parallax_anchor: Vector2 = Vector2.ZERO
```
mapedorr, line 81: Move this above total_frames so all @export variables stay together.

addons/popochiu/engine/objects/prop/popochiu_prop.gd
```diff
@@ -50,6 +50,40 @@ signal obstacle_state_changed(prop: PopochiuProp)
...
+
+var _original_position: Vector2
+@onready var _parallax_enabled: bool = PopochiuUtils.e.settings.parallax_scrolling
+@onready var _parallax_weight := float(_parallax_enabled)
```
mapedorr, line 85: Move this next to `@onready var _navigation_obstacle...` so all the on-ready variables stay together.

addons/popochiu/engine/objects/prop/popochiu_prop.gd
```diff
@@ -164,6 +209,50 @@ func _on_movement_ended() -> void:
...
+		_parallax_tween.tween_property(
+			self,
+			"_parallax_weight",
+			float(_parallax_enabled),
+			PopochiuUtils.e.settings.parallax_scrolling_reset_duration
+		)
+
+	position = _original_position + (parallax_prop_offset * _parallax_weight)
+
+
```
mapedorr, line 255: This function should go inside the `#region Private...` region, not inside the `#region Virtual...`.

addons/popochiu/engine/popochiu.gd
```diff
@@ -532,6 +532,30 @@ func get_current_command_name() -> String:
...
+	if not R.current:
+		return viewport_rect
+
+	var viewport_size := viewport_rect.size
+	var room_size := Vector2(R.current.width, R.current.height)
+	var visible_size := viewport_size.min(room_size)
+	var visible_rect := Rect2((viewport_size - visible_size) * 0.5, visible_size)
+	return visible_rect
+
+
```
mapedorr, line 558: This function should go inside the `#region Public...` region. Despite being called `get_visible_room_rect`, the idea is that this region should contain the `get` and `set` functions for the object's properties. This function, on the other hand, returns a `Rect2`, but isn't tied to any variable.

mapedorr (changes requested):
Hi @bonki . Sorry for the late review on this. For the moment, I just put some comments regarding the code order, but I haven't tested the branch yet. Do you mind sharing some of the assets you used to test the parallax effect? I'll try to test the branch ASAP so we can decide if it can be merged to `develop`.

Issue #558, opened 2026-08-19
Restructure docks and importer to make use of the new Tree control node


stickgrinder:
### Topic

Popochiu 2.x carries a heavy legacy of code from Godot 3.x era. During Godot 4.x lifecycle, they released a new [Tree control node](https://docs.godotengine.org/en/stable/classes/class_tree.html) that basically exposes the same interface they did for the scene tree or other tree-like elements.

It seems the new nodes would support everything already done by hand, althought with a slightly different interface, maybe. This can also be a good thing, in that Popochiu can become even more integrated with Godot in terms of user experience.

### Goals

- Remove a good amount of old code that can be now conflated into standard controls
- Reduce the maintenance weight on the docks and importers
- Learn how to put these controls to good use and if they can be a decent base for a better editor for dialog options (there should be an issue for that)
- Make Popochiu even more integrated with Godot

### Your proposal, in detail

Being completely new to Tree nodes, I propose to work on a POC in a separate branch, rebuilding the main dock's "Main" tab with the new model.

Evaluating this proposal can be done in the light of that effort. The result can be considered throw-away code or be the base to port all the docks, but we'll decide this later.

### Record of decisions taken

Nothing for now, I'll provide a branch on which to test a working version of the main dock.
This issue can be rejected by @mapedorr at any moment, if it's considered useless. In that case I'll stop working on the POC.

stickgrinder:
What started as a POC ended up being the whole thing, plus a new set of SVG icons for the toolbar and Popochiu nodes, mostly provided by Fabiano Caputo.

The open PR is ready to merge. It's pretty huge: maybe testing the thing is better-spent time than reading through all of it.

Issue #533, opened 2026-06-01
Hotspot beacon positioning


drbbgh:
### Benefit description

This serves as preparatory work for #199. Currently, we lack a standard way to place a hotspot beacon that indicates the position of images or animations (beacons) representing interactable objects (`PopochiuClickable`). We will introduce a new `Gizmo2D` to specify this position.

### Solution description

- [x] Add a new `beacon_pos` property to `PopochiuClickable`.
- [x] Add a new `Gizmo2D` to `PopochiuClickable`.
- [x] Add the relative Editor Settings.
- [x] Add a new toolbar visibility button for the new `Gizmo2D`.
- [x] Allow the `Gizmo2D` to be placed on the centroid on user's demand through a toolbar editing button.
- [ ] Provide icons for these buttons.

### Exclusions

This change currently only applies to `PopochiuClickable`. Walkable areas and regions are explicitly excluded. We won't automatically adjust the beacon position if the centroid changes.

### Implications

None

PR #561, opened 2026-08-23
refs #558, #533: new SVG icons for Popochiu GUI

Flaconia:


stickgrinder:
I'm integrating this into my work on tree-bases dock so that the UI shows new icons designed by Fabiano.

Issue #412, opened 2025-07-10
Aseprite Importers: Multiple anims per props and inventory items


stickgrinder:
### Description

Introduce a "Group by prefix" flag in the options for props and inventory items.

At the time of writing, a limitation in Aseprite makes this almost impossible. We need a way to export a list of non-contiguous tags in a single unified spritesheet by Aseprite CLI. At the moment this is not possible.

Leaving this here for future reference.

stickgrinder:
A possible strategy is to use the `--frame-range` parameter, by reading adjacent tags with the same prefix and cutting the file in a different when we detect a group.

Another maybe better strategy would be to create a "group" tag on top of the animations that should actually be a group. Something like this:

<img width="730" height="59" alt="Image" src="https://github.com/user-attachments/assets/ff998d13-a991-4ae9-88d6-542cc3e699bf" />

Documenting this we can make sure that grouping happens as the user intended, no guessing, no black magick.

PR #570, opened 2026-09-17
refs #539: Filesystem churn reduction in Popochiu editor flows

### LINKED ISSUE #539: Deleting a room from the Popochiu dock sometimes causes editor errors
### Bug description

When deleting a room from the Main tab (Remove), the editor can enter an inconsistent state that causes errors. The issue is intermittent: it happens when the room being deleted is currently open in the 2D editor and the "delete folder" checkbox is checked.

###  Steps to reproduce

1. Open a room scene
2. Main Tab -> Room -> Remove
3. Error/Crash

Popochiu should not cause errors or crash.

 - **OS type and version**: Manjaro Linux Kernel 7.0.10
 - **Godot version**: 4.6.3
 - **Plugin version or branch**: 2.1.1

stickgrinder:
refs #539: Reduce the filesystem churn during editor file operations

## Context

#539 reports an intermittent editor crash when deleting a room from the Popochiu dock. Investigation during #412 (summarized in the linked issue comments) showed the editor's crash-prone window is the `EditorFileSystem` scan/preview machinery reacting to file churn — repeatedly triggered by full `scan()` calls inside Popochiu's batch file operations. This PR reduces that churn where it is mechanical and safe. It does not attempt to fix #539's crash itself (still unreproduced outside the reporter's environment — see the discussion in the issue).

## What changed

**Audio row deletion validity fix** (`popochiu_audio_row.gd`, third commit):
- Audio rows are dual-role: cue rows (with `audio_cue` set) and audio file rows (audio files without a cue yet, where `audio_cue` is null). `_delete_from_file_system()` unconditionally dereferenced `audio_cue.audio`, so deleting an unassigned audio file row aborted with a nil access after removing the file, leaving its `.import` orphaned (pre-existing on develop: line unchanged). Now the audio file path is resolved by role and both deleted resources get their `update_file()` bookkeeping.

**Deletion flows** (`popochiu_object_row.gd`, `popochiu_gui_templates_helper.gd`:
- Per-file and per-directory `EditorFileSystem.scan()` calls replaced with `EditorFileSystem.update_file(path)` — which quietly removes the entry, its `.import` companion, the UID, and defers script-class updates (no `EditorProgress` dialog, no frontend rescan).
- One single full `scan()` now runs at the end of `_delete_from_file_system()` instead of ~60 scans for a room the size of `living_room` (56 files + 8 folders). Deletion time dropped from seconds (a scan per file) to ~65 ms in automated runs.

**Audio cue deletion** (`popochiu_audio_row.gd`):
- The trailing `scan()` + `scan_sources()` after the (already correct) `update_file()` call are removed.

**Aseprite import** (`animation_creator_base.gd`):
- Removing the throwaway Aseprite JSON now uses `update_file()` instead of a second full scan. The JSON is already registered by the scan that ran when the spritesheet was imported, so the JSON removal needs no rescan.

## What was deliberately *not* changed

The GUI template **application** flow (`copy_components`) keeps its per-dependency scans. A restructure consolidating them to one scan broke the flow's correctness: newly copied files are invisible to `ResourceLoader.get_resource_uid()` until a full `scan()` registers them (verified: `-1` before, valid ID after), and `_update_dependencies()` skips UID replacements for unresolved UIDs — leaving the generated `gui.tscn` pointing at addon resources instead of the game copies (reproducible head-to-head, 4 addon refs left in the output). The copy flow needs the async/threaded restructure (drafted as a follow-up) — a mechanical scan swap is not enough there.

## Verification

- Room deletion flow automated (scene open in the 2D editor, delete-folder checked — the #539 repro conditions), exercised repeatedly on Godot 4.7.2 and Godot 4.6 stable with `game/` backed up and restored per run: deletion completes correctly (folder gone, data file updated, no script errors), ~65 ms vs multiple seconds with per-file scans.
- GUI template change flow (remove old components + copy `sierra`) before/after head-to-head comparison; the unchanged copy loop keeps producing identical outputs (no addon refs, no invalid UIDs — only the usual run-to-run UID regeneration differences).
- Character Aseprite import exercised after the JSON-removal change: import completes, JSON removed, no leftovers.

**Generated A autoload sanitization** (`popochiu_resources.gd`, `tab_audio.gd`, fourth commit, found while verifying the other changes):
- The A singleton generator declared one variable per cue using the audio file's basename verbatim, so freely-named files (e.g. `mwic_-_Silly_Monkey_2020`) produced invalid identifiers that broke the generated script. `PopochiuResources.get_valid_identifier()` now sanitizes the cue resource names at creation and the generated variable names; applied only to the audio flow since the other singletons derive variables from script names, which are valid identifiers already.

## Verification notes for reviewers

- The audio cue deletion path could not be exercised in a project without audio cues (this repo's test project has none); it is parse-checked and the change removes calls only — the `.import` cleanup still goes through the pre-existing `update_file()`.
- The `scan()` after creation popups and the startup `scan_sources()` were left alone deliberately (one per user action, or editor startup only).


mapedorr (approved):
Awesome. Thanks!

PR #571, opened 2026-09-19
refs #407: Per character inventory

### LINKED ISSUE #407: Per-character inventory management
### Benefit description

Currently the engine allows to switch the player character, ideally supporting a multi-character game like Maniac Mansion, Day of the Tentacle or Thimbleweed park.

The only problem is the inventory is just one, so it gets shared if the user change character, making it impossible to implement a real multi-character game.

### Solution description

All the inventory API must be restructured to take the character into account (defaulting to "player"), both for adding and for "reading" the inventory.
Of course, the inventory will have to be represented in the engine in a way that's different from now, maybe using a dictionary.

**IMPORTANT**: we should strive to leave the public surface of the inventory system as similar to the current one as possible. Breaking changes will have to be properly documented and mapped in a dedicated migration.

mapedorr:
This PR closes #407 by allowing Popochiu to handle each character's inventory and properly update the GUI when players switch `C.player`.

Additionally, it refactors the GUI scripts so they no longer access the inventory component directly. Instead, each component connects to the corresponding signals and updates when items are added to or removed from the inventory, or when they are swapped.

When a game is saved, the inventory of each character is also persisted, ensuring that no data is lost when loading a saved game.


addons/popochiu/engine/objects/gui/components/inventory_grid/inventory_grid.gd
```diff
@@ -45,8 +45,111 @@ func _ready():
...
+	
+	box.get_meta(item.script_name).remove_child(item)
+	box.get_meta(item.script_name).name = EMPTY_SLOT
+	
+	_check_scroll_buttons()
+	
+	await get_tree().process_frame
+
+
+func swap_item(
```
stickgrinder, line 105: All these functions deserve a docblock.

addons/popochiu/engine/objects/prop/popochiu_prop.gd
```diff
@@ -444,12 +444,12 @@ func get_current_animation_position() -> float:
...
 
 #region Private ####################################################################################
-func _on_item_added(item: PopochiuInventoryItem) -> void:
+func _on_item_added(item: PopochiuInventoryItem, _character: PopochiuCharacter) -> void:
 	if item.script_name == link_to_item:
 		disable()
 
 
-func _on_item_removed(item: PopochiuInventoryItem) -> void:
+func _on_item_removed(item: PopochiuInventoryItem, _character: PopochiuCharacter) -> void:
```
stickgrinder, line 452: Why do  we have a `_character` param which is never uses?
If those are virtuals, they are in the wrong region.

mapedorr: It is because the signature of the signals in **PopochiuIInventory** changed in order to handle the item for the target character.

addons/popochiu/engine/objects/gui/components/inventory_grid/inventory_grid.gd
mapedorr: Doing that in my next commit. Thanks!

